PR #20, opened 2023-04-20
Update 8108

Vertiq-Ben:
- Added new client files for 81-08 speed


docs/clients/stopping_handler.rst
```diff
@@ -0,0 +1,121 @@
+Stopping Handler
+----------------
+A Vertiq module is considered stopped when it has been below its stopping speed continuously for some stopping time. 
+Anytime the module’s velocity goes above the stopping speed, it will reset the countdown on the stopping time.
+Vertiq modules execute a disarming process when disarming in order to try and safely come to a stop
```
iq-fred, line 5: You can ditch this stuff about the disarming process, its not stop specific really. Maybe just keep the first two sentences

docs/clients/arming_handler.rst
```diff
@@ -0,0 +1,149 @@
+Arming Handler
+--------------
+Vertiq Advanced Speed modules can support an advanced arming feature,
```
iq-fred, line 3: Can you do a link to the advanced arming section in the Feature Reference Manual section here, so people can see that documentation easily?

docs/clients/arming_handler.rst
```diff
@@ -0,0 +1,149 @@
...
+| Sub ID | Short Name                          | Access          | Data Type | Unit      | Note                                                                                                                                                                                                                                   |
++========+=====================================+=================+===========+===========+========================================================================================================================================================================================================================================+
+| 1      | always_armed                        | get, set, save  | uint8     |           | If True, handler will be automatically armed.                                                                                                                                                                                          |
```
iq-fred, line 115: I may change the description to "If True, handler will automatically arm when the first throttle message is received."

docs/clients/arming_handler.rst
```diff
@@ -0,0 +1,149 @@
...
+|| 11    || manual_arming_throttle_source      || get, set, save || uint8    ||          || The throttle source used as its armed throttle source for manual arming. The default manual throttle source is Unknown. In order to use manual arming, a throttle source must be set first. See row below for each state:             |
+||       ||                                    ||                ||          ||          || 0 = Unknown:                                                                                                                                                                                                                          |
+||       ||                                    ||                ||          ||          || 1 = Hobby: REFER TO Throttle Sources page when merged in docs                                                                                                                                                                         |
```
iq-fred, line 144: Instead of REFER TO, can you make this a real internal reference link? I can show you how to do that if you are unfamiliar. Also this says "when merged in docs", which we should lose now

docs/clients/stow_user_interface.rst
```diff
@@ -0,0 +1,144 @@
...
++--------+---------------------+-----------------+-----------+-------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 3      | sample_zero         | set             |           |                   | Sets the module's current postiion as the zero angle.                                                                                                                                           |
++--------+---------------------+-----------------+-----------+-------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 4      | stow                | set             |           |                   |                                                                                                                                                                                                 |
```
iq-fred, line 123: No notes for this one? Maybe just add something like "Setting this triggers the manual to stow"

iq-fred (changes requested):

docs/clients/coil_temperature_estimator.rst
iq-fred, line 1: The message table here is out of date, compare it to the Python API. For the 8108 we only support from 0 to 13, so you may need to split this into two tables or something and append the second table when we go to add the 40XX documentation, since that supports all of them currently

docs/clients/esc_propeller_input_parser.rst
iq-fred, line 1: This is out of date as well, had some things added to it. Expand message table

iq-fred (changes requested):

iq-fred (approved):

PR #23, opened 2023-04-28
Add regen warning to getting started

iq-fred:


docs/tutorials/testing_with_control_center.rst
```diff
@@ -33,6 +33,10 @@ section of our website, and individual links are provided below for convenience:
 
 Hardware Setup
 ==============
+.. warning:: **If you are using a power supply to power your module, it is possible  to damage or destroy your module with aggressive commands.** This is because Vertiq modules can also act as generators.
```
vertiq-jordan, line 36: I think you should specify what *aggressive commands* are. We know that the biggest problem comes from people stopping hard, but people reading this won't. Even if it is wordier, I think it is best to be as clear as possible about what can cause the issues.

docs/tutorials/testing_with_control_center.rst
```diff
@@ -33,6 +33,10 @@ section of our website, and individual links are provided below for convenience:
 
 Hardware Setup
 ==============
+.. warning:: **If you are using a power supply to power your module, it is possible  to damage or destroy your module with aggressive commands.** This is because Vertiq modules can also act as generators. 
+  Power supplies generally cannot absorb the energy generated by the module, while batteries can, which can lead to dangerous voltage spikes when connected to a power supply.
```
vertiq-jordan, line 37: In general, power supplies, unlike batteries, cannot absorb the energy generated by the module. As such, aggressive commands can lead ...

docs/tutorials/testing_with_control_center.rst
```diff
@@ -33,6 +33,10 @@ section of our website, and individual links are provided below for convenience:
 
 Hardware Setup
 ==============
+.. warning:: **If you are using a power supply to power your module, it is possible  to damage or destroy your module with aggressive commands.** This is because Vertiq modules can also act as generators. 
+  Power supplies generally cannot absorb the energy generated by the module, while batteries can, which can lead to dangerous voltage spikes when connected to a power supply. 
+  To prevent damage to the module when commanding it aggressively on a power supply, it is recommended to turn on the regeneration voltage protection feature.
```
vertiq-jordan, line 38: link to the regen voltage protection feature? Not sure if we have one

iq-fred: Sadly, we do not have any documentation to link to on it. Hopefully we will in the future

vertiq-jordan (approved):

PR #30, opened 2023-05-11
updated matlab examples in iquart_flight_controller_interface and pow…

Vertiq-Ben:
…er_safety. Added new uavcan_node.rst.

- I'm also attempting to provide more detailed values in the Unit field of the message table. These are not reflected in the API json files yet.


docs/clients/iquart_flight_controller_interface.rst
```diff
@@ -119,14 +120,14 @@ Message Table
...
++--------+--------------+----------------+-----------+-----------------+-------------------------------------------------------------------------------------------------------------------+
+| Sub ID | Short Name   | Access         | Data Type | Unit            | Note                                                                                                              |
++========+==============+================+===========+=================+===================================================================================================================+
+| 1      | telemetry    | get            | uint8     | 16 Byte Array   | Returns 16 bytes of telemetry information. See the :ref:`IFCITelemetryData struct<ifcitelemetrydata_note>` notes. |
++--------+--------------+----------------+-----------+-----------------+-------------------------------------------------------------------------------------------------------------------+
+| 2      | throttle_cvi | get, set, save | uint8     | Number [0, 255] | The control value index for the throttle.                                                                         |
```
iq-fred, line 128: Maybe call it index instead of number for these three CVIs?

docs/clients/iquart_flight_controller_interface.rst
```diff
@@ -119,14 +120,14 @@ Message Table
...
-| 4      | y_cvi        | get, set, save | uint8     |      | The control value index value for the y rectangular coordinate.                                                   |
-+--------+--------------+----------------+-----------+------+-------------------------------------------------------------------------------------------------------------------+
++--------+--------------+----------------+-----------+-----------------+-------------------------------------------------------------------------------------------------------------------+
+| Sub ID | Short Name   | Access         | Data Type | Unit            | Note                                                                                                              |
++========+==============+================+===========+=================+===================================================================================================================+
+| 1      | telemetry    | get            | uint8     | 16 Byte Array   | Returns 16 bytes of telemetry information. See the :ref:`IFCITelemetryData struct<ifcitelemetrydata_note>` notes. |
```
iq-fred, line 126: The data type here isn't uint8. I'd say the data type is the 16 Byte Array really and the unit is tough, its really many different units all packed together. Maybe the unit should just be a link to the Telemetry data struct like you have in the note?

docs/clients/power_safety.rst
```diff
@@ -112,36 +111,36 @@ Message Table
...
+| Sub ID | Short Name            | Access         | Data Type | Unit            | Note                                                                                                                                                                                                                                               |
++========+=======================+================+===========+=================+====================================================================================================================================================================================================================================================+
+| 0      | fault_now             | get            | uint8     | Number [0, 255] | The fault(s) that are currently triggering                                                                                                                                                                                                         |
```
iq-fred, line 117: Maybe the unit for fault_now, fault_ever, and fault_latching would be better as "Bitmask". They are numbers, but more specifically they are acting as bitmasks

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
+| Sub ID | Short Name             | Access         | Data Type | Unit             | Note                                                                                                                                                                                                                                                                                                                    |
++========+========================+================+===========+==================+=========================================================================================================================================================================================================================================================================================================================+
+| 0      | uavcan_node_id         | get, set, save | uint32    | Number [1, 127]  | This ID is used by the module to uniquely identify itself on the DroneCAN bus. The module must be power cycled after changing the ID. Each module on the bus must have a unique ID. While the IDs can be a number between 1 and 127, it is recommended to avoid using 1 as this is commonly used by flight controllers. |
```
iq-fred, line 109: Maybe "ID" would be a more specific unit? Not sure

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 1      | uavcan_esc_index       | get, set, save | uint32    | Number [0, 19]   | This is used to identify the module when ESC commands are set on the DroneCAN bus. Each ESC/module on the bus requires a unique index to identify which commands are intended for it. The ESC indexes range from 0 to 19, with the first module starting at index 0.                                                    |
```
iq-fred, line 111: Similarly, maybe index for the unit?

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 2      | zero_behavior          | get, set, save | uint32    | Number [0, 2]    | This determines how the module reacts to receiving a zero setpoint from a DroneCAN Raw Command when DroneCAN is bypassing arming. 0 \= coast, 1 \= Brake, 2 \= Normal Controller                                                                                                                                        |
```
iq-fred, line 113: Have we used enum as a unit before? That might be the right sort of thing to use for zero_behavior

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 3      | last_error_code        | get            | uint8     | Number [0, 8]    | This represents the error code of the last CAN failure. 0 means no error.                                                                                                                                                                                                                                               |
```
iq-fred, line 115: I'm not sure what the unit should be here. Could just be "Error Code" perhaps

iq-fred (changes requested):

docs/clients/iquart_flight_controller_interface.rst
```diff
@@ -87,12 +87,13 @@ A minimal working example for the IQUartFlightControllerInterfaceClient is:
 
 .. code-block:: Matlab
 
-    % Make a communication interface object
-    com = MessageInterface(’COM18’,115200);
-    % Make a IQUartFlightControllerInterfaceClient object with obj_id 0
-    iquartFlightControllerInterface = IQUartFlightControllerInterfaceClient(’com’,com);
-    % Use the IQUartFlightControllerInterfaceClient object
-    pulsingVoltageMode = iquartFlightControllerInterface.get(’telemetry’);
+%   % Make a communication interface object
```
vertiq-jordan, line 90: it's hard to tell from the github version, but do you mean to have these all double commented/the code commented? if people try to copy and paste, then they're just going to get a bunch of commented code

docs/clients/uavcan_node.rst
vertiq-jordan, line 109: The module must be power cycled in order for a new ID to take effect. I think that might be clearer

docs/clients/uavcan_node.rst
vertiq-jordan, line 111: ESC commands are sent instead of set

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
+| 4      | receive_error_counter  | get            | uint8     | Number of errors | This counter increments witch each CAN receive error.                                                                                                                                                                                                                                                                   |
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 5      | transmit_error_counter | get            | uint8     | Number of errors | This counter increments witch each CAN transmit error.                                                                                                                                                                                                                                                                  |
```
vertiq-jordan, line 119: witch typo

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
+| 6      | bus_off_flag           | get            | uint8     | Bool             | This flag indicates if the bus turned itself off due to errors. 1 means the bus is off.                                                                                                                                                                                                                                 |
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 7      | error_passive_flag     | get            | uint8     | Bool             | This flag indicates if the error limit to enter passive mode has been reached. 1 means the module is past the passiv error limit.                                                                                                                                                                                       |
```
vertiq-jordan, line 123: passiv typo

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
+| 8      | error_warning_flag     | get            | uint8     | Bool             | This flag indicates if the error warning limit has been reached. 1 means there has been enough errors to trigger a warning.                                                                                                                                                                                             |
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 9      | telemetry_frequency    | get, set, save | uint32    | Hz               | The frequency with which telemetry will be sent to the DroneCAN bus                                                                                                                                                                                                                                                     |
```
vertiq-jordan, line 127: Add a period at the end

docs/clients/uavcan_node.rst
```diff
@@ -0,0 +1,132 @@
...
+| 10     | bit_rate               | get, set, save | uint32    | Bit/s            | The bitrate used by the DroneCAN bus.                                                                                                                                                                                                                                                                                   |
++--------+------------------------+----------------+-----------+------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| 11     | bypass_arming          | get, set, save | uint8     | Bool             | This setting allows the module to bypass arming with DroneCAN throttle messages. DroneCAN messages will not be impacted by the arming state, will not cause arming state transistions, and will use the zero behavior setting. This is useful for using the 'old-style' DroneCAN support.                               |
```
vertiq-jordan, line 131: 'old-style' doesn't really mean anything to anyone outside of us. Maybe we should put an actual firmware number where this changed? @iq-fred ?

vertiq-jordan (changes requested):
typos and some clarity issues

docs/clients/uavcan_node.rst
iq-fred: I mean you can still use this 'old-style' on the latest firmware if you have bypass arming, so there isn't really a version where it totally went away. And the 8108 still has this set to true by default. So a version is tricky. if the 'old-style' is problematic, we should probably just lose this sentence altogether

iq-fred:
@Vertiq-Ben Don't merge this until the existing development to master PR gets pulled in, so we don't throw more unexpected changes into that one when its almost done.

iq-fred (approved):

vertiq-jordan (approved):

PR #40, opened 2023-08-21
Update units for incorrect Multiturn entries. Re-conform all units to the same formatting until we can change them all.

vertiq-jordan:
updated units for trajectory trajectory_average_speed, trajectory_queue_mode, and ff. removed ..math directive to make formatting consistent. I want to go through at some point soon to add the ..math directive to all units to make them look nicer, but I'd rather not have only a few with it there. The formatting looks wrong that way.

### DIFF
--- a/docs/clients/multi_turn_angle_control.rst
+++ b/docs/clients/multi_turn_angle_control.rst
@@ -164,7 +164,7 @@ Type ID 59 | Multi-turn Angle Controller
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
 | 20     | trajectory_angular_velocity     | get, set       | float     | rad/s             | Final velocity of the trajectory. Defaults to 0.                                                 |
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
-| 21     | trajectory_angular_acceleration | get, set       | float     | .. math:: rad/s^2 | Final acceleration of the trajectory. Defaults to 0.                                             |
+| 21     | trajectory_angular_acceleration | get, set       | float     | rad/s^2           | Final acceleration of the trajectory. Defaults to 0.                                             |
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
 | 22     | trajectory_duration             | set            | float     | s                 | Duration of trajectory. Trajectory is executed or queued once this is sent.                      |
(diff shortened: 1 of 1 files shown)


docs/clients/multi_turn_angle_control.rst
```diff
@@ -164,21 +164,21 @@ Type ID 59 | Multi-turn Angle Controller
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
 | 20     | trajectory_angular_velocity     | get, set       | float     | rad/s             | Final velocity of the trajectory. Defaults to 0.                                                 |
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
-| 21     | trajectory_angular_acceleration | get, set       | float     | .. math:: rad/s^2 | Final acceleration of the trajectory. Defaults to 0.                                             |
+| 21     | trajectory_angular_acceleration | get, set       | float     | rad/s^2           | Final acceleration of the trajectory. Defaults to 0.                                             |
```
vertiq-jordan, line 167: If you look at the docs as they are now, you'll see that the entries with the ..math directive look a lot nicer, but keep the whole table from looking uniform. I want to go and change everything to use it, but until we can do that, we should take it out here so we can have everything looking uniform.

docs/clients/multi_turn_angle_control.rst
```diff
@@ -164,21 +164,21 @@ Type ID 59 | Multi-turn Angle Controller
...
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
-| 27     | trajectory_queue_mode           | get, set, save | int8      | .. math:: m/s^2   | append=0, overwrite=1                                                                            |
+| 27     | trajectory_queue_mode           | get, set, save | int8      | enum              | append=0, overwrite=1                                                                            |
 +--------+---------------------------------+----------------+-----------+-------------------+--------------------------------------------------------------------------------------------------+
-| 29     | ff                              | get, set       | uint32    |                   | Feed forward term                                                                                |
+| 29     | ff                              | get, set       | uint32    | V                 | Feed forward term                                                                                |
```
vertiq-jordan, line 181: This was missing a unit. 
image
ff represents u, the feed forward value in our PID value, aka an added voltage.

Vertiq-Ben (approved):

iq-fred (approved):

iq-matt (approved):

PR #41, opened 2023-08-23
Update descriptions of obs_angular_displacement and zero_angle to make it more clear what each does

vertiq-jordan:
We've gotten user feedback that our multiturn documentation is lacking. This is a first small step in fixing it by updating the descriptions of obs_angular_displacement and zero_angle to be more precise to what they do. 

I've also gone ahead and change the type of table that the client has so that users no longer have to scroll left and right to see all of the information in the table: 
image


docs/clients/multi_turn_angle_control.rst
vertiq-jordan, line 1: The only real changes are on obs_angular_displacement and zero_angle. The rest are formatting changes as a result of changing this table to be a tight-table.

Vertiq-Ben:
It is easier to read when you don't have to scroll horizontally to see the Note field, but now the row heights are different. Is everyone ok with how the row heights changed?

docs/clients/multi_turn_angle_control.rst
```diff
@@ -118,69 +118,74 @@ Message Table
...
+		+--------+---------------------------------+----------------+-----------+-----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+		| 31     | zero_angle                      | get, set, save | float     | rad       | The encoder position the module considers to be 0 radians. Since this is an encoder position, zero_angle is constrained to [-pi, pi]. Unless obs_angular_displacement is set explicitly, this is the position regarded as 0 radians, and all displacements are measured in comparison to this point.                                               |
```
iq-fred, line 188: This might be slightly confusing wording in the second part, regarding the "Unless obs_angular_displacement is set explicitly, this is the position regarded as 0 radians". This angle is always considered 0 radians in terms of angle commands, but it may not be 0 radiands of displacment if you set obs_angular_displacement. So maybe just add that "this is the position regarded as 0 radians of displacement", and maybe specify that angle commands ("ctrl_angle") they will always be referenced to this? Unless I'm wrong about that stuff, it can be hard to keep it all straight

iq-fred (changes requested):

docs/clients/multi_turn_angle_control.rst
```diff
@@ -118,69 +118,74 @@ Message Table
...
+		+--------+---------------------------------+----------------+-----------+-----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+		| 8      | timeout                         | get, set, save | float     | s         | The controller must receive a message within this time otherwise it is set to coast mode                                                                                                                                                                                                                                                           |
```
Vertiq-Ben, line 144: Add period at the end

Vertiq-Ben (changes requested):
Some Notes fields are missing a period

docs/clients/multi_turn_angle_control.rst
vertiq-jordan: Just did a test to confirm on my desk. Here's what I did:
1. Observed angle was 2.6
2. Set zero angle with the sample button. Observed angle now 0
3. manually turn the motor by pi. observed angle pi. set ctrl_angle to 0. it goes to 0.
4. Coast the motor. manually turn to pi. set observed angle to 0. set ctrl_angle to 0. the motor locks right where it is, and does not go back to the originall set zero angle

vertiq-jordan: So, it's not always regarded as 0 displacement, and ctrl_angle isn't always in reference to this. the ctrl_angle is in reference to whatever is currently 0 displacement. To start, this is zero_angle, but that's not always true

docs/clients/multi_turn_angle_control.rst
vertiq-jordan: I think we inherited this from whenever the table was written, so the inconsistencies make sense. I'll fix them

iq-fred (approved):

iq-matt (approved):

PR #50, opened 2023-12-15
Bugfix/update naming for business

vertiq-jordan:
the business team requested some changes to RTD. One, to make sure that all module names have a dash in them (81-08 instead of 8108). The second was to fix the side bar such that we do not include any KVs.

PR #51, opened 2024-01-17
Bugfix/update naming for business

vertiq-jordan:
the business team asked that we add dashes in all of the motor names, so I went through and did it.


iq-fred (approved):

iq-fred (changes requested):
This is a duplicate of this PR which has comments on it already: https://github.com/iq-motion-control/iq-module-communication-doc/pull/50

PR #58, opened 2024-02-06
Feature/introduce client table reference

vertiq-jordan:
This is part of phase one of our overhaul of RTD. It introduces a new Client Reference Table that stores all of the client tables that exist within the module pages. The idea is that in the future, we would just add links to all of the applicable client tables people could use. A reminder of our discussion: https://docs.google.com/document/d/1a7MV8iF3pcxMVqp392K34cmCaN4RGIf59y9ZQA3-6BU/edit#heading=h.s25xt7qvyr4c. 

If you look at the module pages now, you'll see that every client has an API example with it. I have chosen to leave those off for now, as I want to push the API pages to be better overall, and examples for each client would be useless and redundant.


docs/clients/power_safety.rst
vertiq-jordan, line 1: while working on this, I found a typo in the power safety client table. It is type 84 not 83. I've fixed it here as well as the new reference page.

Vertiq-Ben (approved):

vertiq-jordan:
> I like that the Client Table Reference page has its own tab on the sidebar. Do we plan on adding examples or linking to the existing examples for each client on this page?

I think for the API examples in general, I would like to overhaul them with new, clearer ones. They'd link to here. I don't plan on changing any of the examples throughout the docs right now though.

docs/client_reference/client_reference_table.rst
iq-fred, line 1: My major issue with this is that unlike the existing module files, this duplicates all of the tables from the clients directory. So now we have the same table information in two places, which is rather unsettling. It would be better if this could just include the tables from clients or something like we do with existing module pages to avoid duplication.

iq-fred (changes requested):

docs/client_reference/client_reference_table.rst
vertiq-jordan: done

iq-fred, line 1: Might be worth being a bit more specific and calling this the "IQUART Client Reference Table" or something, to make it clear this is specifically for IQUART.

iq-fred (approved):

PR #68, opened 2024-04-12
Feature/add control center manual

vertiq-jordan:
This PR replaces the current "Testing speed modules with IQ Control Center" tutorial with a full blown guide on all things control center.

You can see the built docs here: https://iqmotion.readthedocs.io/en/feature-add_control_center_manual/index.html


iq-fred:
@vertiq-jordan You're going straight to master instead of develop? Was that intentional?

vertiq-jordan:
fixed it

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+
+.. image:: ../_static/control_center_pics/flash_btn.png
+
+.. note::
+    If you’ve selected a zip archive not meant for your module, you will be shown a warning, and will not be allowed to flash that particular file as it can cause serious damage to your module.
+
+Recovering a Module from a Failed Flash
+===========================================
+.. note::
+    Recovery mode is only available on the 40-XX and 81-XX families
```
iq-fred, line 640: It may be easier for the future to say what things this isn't available on, namely the 2306 and Fortiq I think? And some customs, but they are their whole own thing, we're focused on stocks here. Otherwise we need to update this for the 60, 110, etc.

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+This commands the module to try and maintain a specific velocity in rad/s, using closed loop PID control. A positive velocity always goes in the counterclockwise 
+direction, and a negative velocity always goes clockwise. To learn more about velocity and voltage control 
+see :ref:`manual_velocity_control_mechanisms`
+
+Voltage
+=============
+This commands the module to a specific drive voltage. A positive voltage always goes in the counterclockwise direction, and a negative voltage always goes clockwise.
+
+ESC Input
+=============
```
iq-fred, line 573: Also, this assumes a speed module. Servos won't have ESC input. I think you were trying to be style agnostic on this page.

You may want to move this out to a getting started with speed page, which I think you planned on doing at some point in the future but I'm not sure. Can just have things shared by all modules like voltage, velocity, PWM. Or at least specify this is only for speed modules, but then it would also seem like you should talk about the servo input parser testing for servos, so it may get a bit wonky

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+command, and click the set arrow 10 times**. On the 10th time you click the set arrow, the module should play its 2 note arming song, and begin spinning. 
+Now that the module is armed, any new throttle commands sent using the ESC Input parameter will change how it is spinning. For example, setting ESC 
+Input to 0.5 should set the module to the same speed as setting the Voltage parameter to 2.5V. This is because the module is in Voltage mode with a 
+Max Volts of 5V, so a 50% throttle command will set the module to spin with a drive voltage of 2.5V. For more information on how to configure 
+the module to properly interpret throttle commands, see the Throttle Commands and Modes section of the Feature Reference Manual. For more information on 
+arming and disarming the module, refer to the Advanced Arming section of the Feature Reference Manual.
+
+Try it Yourself
+==================
+Try out some of these parameters and observe how the module spins. For example, try setting the Voltage to 2.5V, then set the ESC Input to 0.5.
```
iq-fred, line 588: This is copied over from the old page, but it doesn't make sense anymore. On that page, I had stepped through a specific setup example with a certain direction, 2D/3D FC mode, and max volts, which made these make sense. There's no example setup prior to this here though, so these commands wouldn't actually do what they say we would unless someone happened to set it up perfectly. It also assumes a speed module.

Having an example here is good, but you may want to re-focus it on common things like Voltage and Velocity instead of ESC Input for this page. This kind of example requires more specificity

The previous page you took this from was specifically focused on walking you through a speed module setup, which isn't really the case anymore with this page. This specific example may make more sense on a getting started with speed page. Or, if you really want to keep it as it is you'd need to at least specify what your direction and max volts configurations are, and specify you are using a speed module

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+    Before using the Control Center's testing functionality, we recommend that you increase your module's timeout parameter to 1.5s. 
+    This parameter is available through the tuning tab. If you do not increase the timeout value, your module may fail to spin as expected.
+
+.. warning::
+    Please remove all propellers before interacting with any testing parameters. Failure to do so is dangerous.
+
+In order to set a testing parameter, simply enter a value into the spin box, and hit the down arrow set button. Parameters such as Coast do not have a spin box, 
+and only require that you press the set button to trigger the behavior.
+
+The testing tab for a Veritq 23-06 2200Kv looks as follows:
```
iq-fred, line 540: This says "Veritq" instead of "Vertiq"

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+.. note::
+    If your installer does not open, please perform the following steps.
+
+    a. Open a new terminal window
+    b. Navigate to the location storing the Control Center Installer
+
+    .. image:: ../_static/control_center_pics/linux/linux_terminal_location.png
+
+    c. Use ./ to run the installer
+    d. If you see the following error You MUST run `sudo apt-get install libxcb-xinerama0`
```
iq-fred, line 97: Any idea what this thing is for or why people might need to do this?

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+
+* The :red:`Tabs` section lists the different tabs of the Control Center that are available when connected to a module. Each tab offers access to different configuration parameters and functionality. The settings in these tabs will be covered in greater detail later.
+
+* The :green:`Information` section lists information about the Control Center and connected module, if there is one. When there is no module connected, this only lists the Control Center version. When a module is connected, this will also list its Firmware Version, Firmware Style, Hardware Style, Bootloader Version, and Upgrader Version. Check this information to make sure you are using the proper hardware with the latest firmware. You can check for new firmware under the Modules section of the Vertiq website.
+
+* The :blue:`Module Connections` section is used to connect to modules. You can select the baud rate and the serial port to use for the module you want to connect to. The serial port is determined by the name of the serial port used by your USB-to-UART converter. See the <link to the right thing> section for more details on how to determine that port name.
```
iq-fred, line 235: This still says "<link to the right thing>" instead of having a link

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+Once you have selected your target baud rate and your USB-to-UART device's serial port, you are ready to connect with your module. 
+
+Connect your module to the USB-to-UART device as described :ref:`above <connection_guide>`, and power on your module. 
+
+Click the connect button in order to open serial communication between the Control Center and your module. 
+
+If you successfully found and communicated with a module, you will see its information appear in the :ref:`Information section <information_box>` in the bottom left. 
+You will also see the :ref:`Module Connections <module_connections>` box list the number of detected modules as non-zero. 
+
+The CONNECT button now presents DISCONNECT.
```
iq-fred, line 308: The phrasing is a bit odd here. Maybe "The CONNECT button will now change to the DISCONNECT button"

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+Connection Information Example
+===============================
+The following sections demonstrate how the :ref:`Control Center's GUI <gui_overview>` is filled in after connecting a module.
+
+.. _information_box:
+
+Information
+--------------
+In the bottom left corner of the Control Center is information about the currently connected module. It provides information about the module's 
+firmware, hardware, bootloader and upgrade versions, and a link to your module's id.vertiq.co.
```
iq-fred, line 325: You could make id.vertiq.co an actual link here for people who don't know what that is

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+Detect
+^^^^^^^^^^^^^^^^^^^^^^^
+Clicking DETECT reruns the module detection process. This allows you to connect modules to the Control Center after already having run the initial connection.
+
+For example, say you've connected one module with Module ID 1
+
+.. image:: ../_static/control_center_pics/one_module_connected.png
+
+Now, you would like to connect a second module to the bus, and configure it via the Control Center. All you have to do is 
+power it on, plug it in to the serial bus, and hit detect.
```
iq-fred, line 356: To be consistent, this should probably be DETECT

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+Vertiq 23-06 2200Kv module with speed firmware with an APC 6x4 propeller. In order to set these parameters, simply select the file that you would 
+like to use, then click Set.
+
+Using User Generated Default Settings
+-----------------------------------------
+User generated default settings allow you to easily share identical module configurations between multiple modules. For example, if you have parameters 
+that work for your module configuration (gains, modes, baud rate, etc.) that should be set on all of your modules, you would want to use a custom defaults file.
+
+.. note::
+    User generated Defaults files include the Module ID parameter. This means that setting up modules with custom Default files will result in collisions on
```
iq-fred, line 496: Not sure if the d in Defaults should be capitalized here, we don't capitalize it in other places we say defaults file

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+
+******************************************
+Testing Modules with the Control Center
+******************************************
+The Control Center provides a basic suite of testing options to ensure that your module is functioning properly. These options are available in the testing tab, 
+available on the left side.
+
+.. note::
+    Before using the Control Center's testing functionality, we recommend that you increase your module's timeout parameter to 1.5s. 
+    This parameter is available through the tuning tab. If you do not increase the timeout value, your module may fail to spin as expected.
```
iq-fred, line 532: You may want to mention that people might want to consider setting it back to a lower value when they are done depending on what they really want the timeout to be.

Also may be worth explaining that the reason for this is because the Control Center only heartbeats once a second, so if your timeout is less than 1s your module will timeout

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+The testing tab for a Veritq 23-06 2200Kv looks as follows:
+
+.. image:: ../_static/control_center_pics/23_testing_tab.png
+    :scale: 50%
+
+Brake
+=========
+This commands the module to brake, which shorts the motor phases, which slows the motor down by dissipating energy in the motor.
+
+.. warning::
```
iq-fred, line 549: This may make sense to move up to the testing introduction section instead of just living with the brake stuff. Its not really brake specific

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+    If you are using a power supply to power your module, it is possible to damage or destroy your module with aggressive commands, e.g. quickly switching from spinning at full speed to stopping. 
+    This is because Vertiq modules can also act as generators. In general, power supplies, unlike batteries, cannot absorb the energy generated by the module. As such, aggressive commands can lead to dangerous 
+    voltage spikes when connected to a power supply. To prevent damage to the module when commanding it aggressively on a power supply, it is recommended to turn on the regeneration voltage protection feature.
+
+Coast
+=============
+Commanding a coast means that the module allows its rotor to slowly decelerate and come to a stop. It is a safe way to stop the module when testing since it does not cause the module to rapidly decelerate.
```
iq-fred, line 556: Consider clarifying that it allows to slowly decelerate by disconnecting all the phases and letting it spin freely, its not like its intentionally driving it gradually to a stop. The wording now could be a bit ambiguous for someone not in the know

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,737 @@
...
+controller, more setup is needed. Refer to the PWM and DSHOT Control with a Flight Controller and the DroneCAN Integration with a PX4 Flight Controller 
+for more details on integrating with flight controllers.
+
+*******************************************
+Updating Firmware with the Control Center
+*******************************************
+Flashing new firmware onto your module is done through the FIRMWARE tab in the Control Center, available on the left side.
+
+All module firmware can be found on your product's page on `Vertiq's website <https://www.vertiq.co/>`_.
+Screen Capture Example
```
iq-fred, line 600: I think Screen Capture Example was meant to be a heading, but it's just regular text so it kind of comes out of nowhere right now

iq-fred (changes requested):

docs/control_center_docs/control_center_start_guide.rst
vertiq-jordan: I think it's a graphics/window manager. Doesn't ship by default on linux machines. 

"provides an interface to the X Window System protocol, which replaces the current Xlib interface"

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+On modules with exposed serial pads, we recommend using cabling with one end set into a Servo JR female connector, and the other 
+end soldered directly to the module's pads. On the 81-08 G1, we recommend using a cable with a Servo JR female connector on both sides.
+
+Using Servo JR connectors makes it easy to connect with, and disconnect from the USB-to-UART device that you are using to connect with the Control Center. 
+
+Single Module Wiring
+---------------------------
+To communicate with a single module, simply connect the Servo JR female connector to the FTDI.
+
+**Please ensure that your USB-to-UART device's TX port is connected to your module's RX port, and your module's TX to the USB-to-UART device's RX.**
```
iq-fred, line 199: In the first part of this, you say TX port and RX port, then in the second half its just TX and RX, no port. May want to pick one for consistency

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+Multiple Module Wiring
+---------------------------
+If you plan on connecting several modules to the Control Center at once, you must connect all of the modules' serial ports together. 
+How you do this is up to you, but you must ensure that the resulting wiring with your USB-to-UART device looks as follows:
+
+.. image:: ../_static/control_center_pics/multi_module_serial.png
+
+We highly recommend connecting your modules together with a non-permanent, easily detachable method. Some configurations through the Control Center 
+must be performed on independently connected modules before multiple can be connected at once. So, having a quick way to attach/detach modules will 
+make the process much easier. A simple way to accomplish this is with 3-pin, 100-mil header pins and a breadboard.
```
iq-fred, line 215: This is mostly relevant for the initial module setup, once you've got everything setup, you can switch to a more permanent solution. People probably don't want a bread-board on their final drone. Maybe worth mentioning this is mostly important for initial setup so people don't think our long-term wiring recommendation involves a bread-board. Or cut down on specifics this section and just point out that there is some initial individual module setup so we recommend taking that into account for your initial wiring setup.

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+
+* The :purple:`Messages` section displays information about the Control Center that can be useful for understanding what state the Control Center is in and debugging connection problems. Check here if you are having connection issues for feedback.
+
+**********************************
+Connecting With your Module(s)
+**********************************
+
+.. _multi_module_config:
+
+Required Configuration Before Connecting with Multiple Modules
```
iq-fred, line 247: One thing that occurred to me, may be worth putting a note or warning here warning that some older firmwares may not support this. Like, the old 2306 servo firmware, pretty sure that is too old to support this feature

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+The names that appear in this box are OS dependent. For example, on Linux, you will see devices with the name “/dev/ttySX” where X is the value of the port.
+
+Connect Button
+--------------------
+
+Once you have selected your target baud rate and your USB-to-UART device's serial port, you are ready to connect with your module. 
+
+Connect your module to the USB-to-UART device as described :ref:`above <connection_guide>`, and power on your module. 
+
+Click the connect button in order to open serial communication between the Control Center and your module.
```
iq-fred, line 305: Should be CONNECT button if you want to go with the all-caps for buttons

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+
+If you successfully found and communicated with a module, you will see its information appear in the :ref:`Information section <information_box>` in the bottom left. 
+You will also see the :ref:`Module Connections <module_connections>` box list the number of detected modules as non-zero. 
+
+The CONNECT button now changes into the DISCONNECT button.
+
+Disconnect Button
+--------------------
+
+Clicking the disconnect button closes the serial connection between the Control Center and your connected USB-to-UART device.
```
iq-fred, line 315: Same thing about capitalization

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+
+.. note::
+    The configuration parameters presented for each module family and firmware style vary. Not all parameters shown in this example will be available for all modules and firmware styles.
+
+General
+-----------
+The general tab provides access to some more basic module configuration parameters. Parameters here help configure attributes necessary for properly controlling 
+your module with your desired behaviors.
+
+For example, the general tab is where you will find the :ref:`Motor Direction parameter <throttle_direction>`.
```
iq-fred, line 463: You'll only find it there for speed firmware, worth specifying that if this is meant to be general cross-firmware directions

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+
+.. warning::
+    If you are using a power supply to power your module, it is possible to damage or destroy your module with aggressive commands, e.g. quickly switching from spinning at full speed to stopping. 
+    This is because Vertiq modules can also act as generators. In general, power supplies, unlike batteries, cannot absorb the energy generated by the module. As such, aggressive commands can lead to dangerous 
+    voltage spikes when connected to a power supply. To prevent damage to the module when commanding it aggressively on a power supply, it is recommended to turn on the regeneration voltage protection feature.
+
+In order to set a testing parameter, simply enter a value into the spin box, and hit the down arrow set button. Parameters such as Coast do not have a spin box, 
+and only require that you press the set button to trigger the behavior.
+
+The testing tab for a Vertiq 23-06 2200Kv looks as follows:
```
iq-fred, line 547: May be worth specifying this is with speed firmware (we do have other things like pulsing available for the 2360 2200kv), and that the user may see more or less options here depending on their firmware.

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+
+.. image:: ../_static/control_center_pics/23_testing_tab.png
+    :scale: 50%
+
```
iq-fred, line 551: A little bit of intro here to say that you are going to describe some common testing parameters across firmwares is useful to motivate and explain these sections below, since they aren't a 1-to-1 with your picture anymore. That's good that they aren't, since ESC Input is speed-specific, I think its just helpful to give an introduction explaining why you are talking about this specific subset of the testing parameters

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -0,0 +1,740 @@
...
+           
+           * Flash App: Updates only the application code, which is the section of code that runs during normal operation.
+
+           * Flash Boot: Updates the proprietary bootloader.
+
+           * Flash Upgrade: Updates the proprietary upgrade firmware.
+
+           * Flash Combined: Updates all three sections.
+
+        **If you are unsure of which type of update to perform, it is recommended to select Flash Combined.**
```
iq-fred, line 627: I might actually say that if they have the app one as an option, hit that, that's the safest. Otherwise, select combined.

iq-fred (changes requested):

iq-fred (approved):

PR #75, opened 2024-06-20
Add 60 Fam Page, Reset to Defaults Info, and Regen Info

vertiq-jordan:
Adds 3 new sections: 
1. 60 family page
2. Reset to defaults information
3. Regen information

You can find the built docs here: https://iqmotion.readthedocs.io/en/feature-reset_defaults_60_fam_regen/tutorials/protect_against_regen.html


docs/modules/advanced_speed_servo_info.rst
```diff
@@ -44,6 +44,8 @@ Supported Tutorials
         * :ref:`dronecan_fc_tutorial`
         * :ref:`fc_telemetry_tutorial`
         * :ref:`motor_noise_debugging`
+        * :ref:`vibration_and_jittering`
```
vertiq-jordan, line 47: added missing tutorials

docs/modules/basic_speed_servo_info.rst
```diff
@@ -36,6 +36,8 @@ Supported Tutorials
         * :ref:`hobby_fc_tutorial`
         * :ref:`hobby_calibration_tutorial`
         * :ref:`fc_telemetry_tutorial`
+        * :ref:`vibration_and_jittering`
```
vertiq-jordan, line 39: added missing tutorials

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+
+What are Regenerative Voltage Spikes
+=======================================
+Regeneration voltage occurs when the commanded module voltage is less than the module's back-EMF voltage. When this occurs, the motor's kinetic 
+energy is converted directly to electrical energy, and produces a current flowing out of the motor and into the power source.
+
+How Can Regeneration Affect Your Module
+=========================================
+If the module's power source can absorb the regenerated current without allowing the voltage to spike, regeneration protection isn't needed. This is often true with a battery as the power source.
+When connected to a less than constant voltage source, like a benchtop power supply, the overall system voltage will increase during regeneration.
```
iq-fred, line 18: I'm not quite sure I understand the "less than constant" bit of this. A battery isn't a constant voltage source either, its voltage is dipping and changing all the time based on the current you draw out of it and is it discharges. Both power supplies and batteries aren't really constant, though they do have different behaviors. I would probably change this to "When connected to a power source that cannot absorb this regenerated current, which is generally the case with benchtop power supplies, the output voltage of the source will often increase during regeneration."

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+What are Regenerative Voltage Spikes
+=======================================
+Regeneration voltage occurs when the commanded module voltage is less than the module's back-EMF voltage. When this occurs, the motor's kinetic 
+energy is converted directly to electrical energy, and produces a current flowing out of the motor and into the power source.
+
+How Can Regeneration Affect Your Module
+=========================================
+If the module's power source can absorb the regenerated current without allowing the voltage to spike, regeneration protection isn't needed. This is often true with a battery as the power source.
+When connected to a less than constant voltage source, like a benchtop power supply, the overall system voltage will increase during regeneration. 
+These spikes can result in voltages above the module's and your supply's maximum rated operating voltage. When this happens, you can permanently damage the module, the power supply, or both.
```
iq-fred, line 19: Repeated sentence here, the "When this happens...." is twice in a row

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+How Can Regeneration Affect Your Module
+=========================================
+If the module's power source can absorb the regenerated current without allowing the voltage to spike, regeneration protection isn't needed. This is often true with a battery as the power source.
+When connected to a less than constant voltage source, like a benchtop power supply, the overall system voltage will increase during regeneration. 
+These spikes can result in voltages above the module's and your supply's maximum rated operating voltage. When this happens, you can permanently damage the module, the power supply, or both. 
+When this happens, you can permanently damage the module, the power supply, or both.
+
+Protecting Against Dangerous Regeneration Voltage Spikes
+===========================================================
+Regenerative voltage spikes pose a serious threat to both your modules and your power supply. As such, all Vertiq modules are able to self-limit the amount of
```
iq-fred, line 24: Maybe don't be quite so alarmist, they can be fine if they are kept under control. Maybe more like, "Regenerative voltage spikes can pose a threat to your modules and power supply if not properly controlled"

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+=========================================
+If the module's power source can absorb the regenerated current without allowing the voltage to spike, regeneration protection isn't needed. This is often true with a battery as the power source.
+When connected to a less than constant voltage source, like a benchtop power supply, the overall system voltage will increase during regeneration. 
+These spikes can result in voltages above the module's and your supply's maximum rated operating voltage. When this happens, you can permanently damage the module, the power supply, or both. 
+When this happens, you can permanently damage the module, the power supply, or both.
+
+Protecting Against Dangerous Regeneration Voltage Spikes
+===========================================================
+Regenerative voltage spikes pose a serious threat to both your modules and your power supply. As such, all Vertiq modules are able to self-limit the amount of 
+regeneration voltage produced. Vertiq modules have two main parameters to protect against dangerous regeneration voltage spikes. They are available through IQ Control Center's advanced tab as
```
iq-fred, line 25: It's really the supply that's "producing" the voltage spike, may be good to clarify that with something like "As such, all Vertiq modules can limit their regeneration current to limit the magnitude of voltage spikes on power supplies", or something like that. My reasoning is that often customers believe regeneration is a magical feature of the motors that we could just flip off, when really its down to the power supplies and their inability to sink current, and I just want to discourage that kind of thinking.

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+regeneration voltage produced. Vertiq modules have two main parameters to protect against dangerous regeneration voltage spikes. They are available through IQ Control Center's advanced tab as 
+*Volts Limit* and *Volts Limit Starting Voltage*.
+
+For example, by default on Vertiq 81-XX family modules (rated to 14S), you will see:
+
+.. image:: ../_static/control_center_pics/regen_params.png
+        :alt: Regeneration Parameters in IQ Control Center
+
+In short, these two values work together in order to limit the rate at which your module will slow down or apply a braking force. In doing so, the module limits the amount of 
+negative supply current produced, reducing the produced regenerative voltage.
```
iq-fred, line 34: In the same vein as my comment above about clarifying the power supplies role in this, I may just tweak this last bit to be "...reducing the voltage spike produced by the power supply

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. note::
+    These configuration parameters define **system voltages**. For example, if your module has a maximum voltage rating of 14S (or 58.8V), but your power supply has a 
+    maximum rating of 30V, the values of Volts Limit Starting Voltage and Volts Limit must be set according to the absolute 30V maximum.
```
iq-fred, line 38: May add on the end here "the absolute 30V maximum of the power supply" just for extra specificity

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. _protect_against_regen:
+
+*************************************************************************
+Configuring Your Module to Protect Against Regeneration Voltage Spikes
+*************************************************************************
+
+What are Regenerative Voltage Spikes
+=======================================
+Regeneration voltage occurs when the commanded module voltage is less than the module's back-EMF voltage. When this occurs, the motor's kinetic
```
iq-fred, line 12: What sort of customer situations does this happen in? Our customers are often not so in the know about actual motor physics stuff like Back-EMF and all. So it may be good to give some ideas here of when this would happen in terms of what they think of. E.g. When the motor is slowing down or braking. And a harder step down will generally produce a more extreme current.

iq-fred (changes requested):

docs/tutorials/protect_against_regen.rst
vertiq-jordan: Hrm. Ok, I'm fine with that. This was just stolen directly from Mike's suggestion to change what I had initially.

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,48 @@
...
+When connected to a power source that cannot absorb this regenerated current, which is generally the case with benchtop power supplies, the output voltage of the source will often increase during regeneration. 
+These spikes can result in voltages above the module's and your supply's maximum rated operating voltage. When this happens, you can permanently damage the module, the power supply, or both.
+
+Protecting Against Dangerous Regeneration Voltage Spikes
+===========================================================
+Regenerative voltage spikes can pose a threat to your modules and power supply if not properly controlled. As such, all Vertiq modules can limit their regeneration current to limit the magnitude of voltage spikes on power supplies. 
+Vertiq modules have two main parameters to protect against dangerous regeneration voltage spikes. They are available through IQ Control Center's advanced tab as *Volts Limit* and *Volts Limit Starting Voltage*.
+
+For example, by default on Vertiq 81-XX family modules (rated to 14S), you will see:
```
iq-fred, line 29: You may want to specify you are talking about 81G2 modules here. The G2s are rated to 14S, but the G1s are only rated to 12S, wouldn't want to give people the wrong idea

iq-fred (changes requested):

iq-fred (approved):

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,48 @@
...
+When connected to a power source that cannot absorb this regenerated current, which is generally the case with benchtop power supplies, the output voltage of the source will often increase during regeneration. 
+These spikes can result in voltages above the module's and your supply's maximum rated operating voltage. When this happens, you can permanently damage the module, the power supply, or both.
+
+Protecting Against Dangerous Regeneration Voltage Spikes
+===========================================================
+Regenerative voltage spikes can pose a threat to your modules and power supply if not properly controlled. As such, all Vertiq modules can limit their regeneration current to limit the magnitude of voltage spikes on power supplies. 
+Vertiq modules have two main parameters to protect against dangerous regeneration voltage spikes. They are available through IQ Control Center's advanced tab as *Volts Limit* and *Volts Limit Starting Voltage*.
+
+For example, by default on Vertiq 81-XX Gen 2 family modules (rated to 14S), you will see:
```
iq-matt, line 29: I think you should refrain from using concrete examples.  If the specs change, you'll have to come back and change this example, which we almost certainly won't remember to do.  

Instead, you can just say "For example, for our modules rated to 14S, you would see:"

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,48 @@
...
+        :alt: Regeneration Parameters in IQ Control Center
+
+In short, these two values work together in order to limit the rate at which your module will slow down or apply a braking force. In doing so, the module limits the amount of 
+negative supply current produced, reducing the voltage spike produced by the power supply.
+
+.. note::
+    These configuration parameters define **system voltages**. For example, if your module has a maximum voltage rating of 14S (or 58.8V), but your power supply has a 
+    maximum rating of 30V, the values of Volts Limit Starting Voltage and Volts Limit must be set according to the power supply's absolute 30V maximum.
+
+    As another example, suppose you want to power your Vertiq 60-08 (rated up to 12S) with a 6S power supply. In this case, you would set your voltage limits based on the
```
iq-matt, line 41: Same as my above comment.  Don't put a number for the 60-08 specifically.  The 60 may actually become 14S in the future.  Just use an anonymous example here instead.

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,48 @@
...
+    6S power supply.
+
+    By default, Vertiq modules have these values set to protect the module's circuits while operating with the supplied voltage at its maximum rating. In order to protect your power supply and module, 
+    please ensure that your configured voltage limits are set according to the **system's lowest maximum voltage rating**.
+
+* **Volts Limit Starting Voltage** defines the value (in volts) at which the module will begin checking if any limits on regeneration need to be applied. The value of Volts Limit Starting Voltage must always be at least 1V below Volts Limit.
```
iq-matt, line 47: "Setting this value to 0V would always enable checking.  Setting this value to just below Volts Limit would still enable the maximum breaking under normal conditions, and only enable the protection once the voltage spikes."  

Or something like that.

docs/tutorials/protect_against_regen.rst
```diff
@@ -0,0 +1,48 @@
...
+* **Volts Limit** defines the absolute maximum supply voltage the module will apply. In some extreme instances, your module may apply a voltage higher than the configured Volts Limit. In order to handle this properly, ensure that there is a buffer between the absolute maximum allowable system voltage and the value of Volts Limit. For example, if your system can handle a maximum of 30V, your Volts Limit may be 27V.
```
iq-matt, line 48: Either in this description or in a dedicated WARNING block, you should mention:
"Never set Volts Limit below the maximum supply voltage.  For example, if you're using a 6S battery with a maximum voltage of 25.2V, never set Volts Limit to 25.2V or lower.  Violating this rule may cause the motor to run away to its maximum speed."

Or something like that.

iq-matt (changes requested):
I left a couple inline comments in the regen protection rst.

vertiq-jordan:
> I left a couple inline comments in the regen protection rst.

Updated.

PR #77, opened 2024-07-02
New API Docs/Update Python Docs

vertiq-jordan:


docs/index.rst
```diff
@@ -6,13 +6,13 @@
...
 .. toctree::
    :caption: API Documentation
-   :titlesonly:
    :hidden:
 
-   langs/python
-   langs/arduino
-   langs/cpp
-   langs/matlab
+   api_docs/getting_started_with_apis
```
vertiq-jordan, line 11: for some reason we set this to titles only which meant no subsections. That was weird, and I needed them, so I've changed it. I also renamed the langs folder to api_docs to be clear.

iq-fred:
@vertiq-jordan If this is only part 1 of the API updates, do you want this to go into master like it is now or do you want it to go into develop, or even a separate API develop where you can stage your changes?

vertiq-jordan:
> @vertiq-jordan If this is only part 1 of the API updates, do you want this to go into master like it is now or do you want it to go into develop, or even a separate API develop where you can stage your changes?

Yea, I'll make a staging branch to store these changes

docs/api_docs/getting_started_with_apis.rst
```diff
@@ -0,0 +1,39 @@
...
+Getting Started with Vertiq's APIs
+***************************************
+
+What is the Purpose of Vertiq's APIs
+===========================================
+Vertiq's APIs provide a simple mechanism for communicating with your modules via the IQUART protocol. IQUART is Vertiq's proprietary serial 
+protocol meant to allow configuration and control of all Vertiq modules. To learn more about IQUART, please see :ref:`the protocol's documentation <uart_messaging>`. 
+Our APIs provide support for Python, Matlab, C++, and Arduino. These options provide flexibility to work easily on both PC-based and embedded solutions.
+
+If you intend to use the API specifically in a PC based environment, we recommend using either the Python or Matlab APIs.
```
iq-fred, line 16: This may be a bit limiting. You can use the C++ one on the PC, and it can actually be quite a bit faster than the Python one in my experience, so there are cases where people may want to do that. I'd say either ditch this altogether, or say something closer to "Python or Matlab are generally easier to get started with, though it is also possible to use C++ on the PC"

docs/api_docs/getting_started_with_apis.rst
```diff
@@ -0,0 +1,39 @@
...
+
+If you intend to use the API in an embedded system, we recommend using the C++ or Arduino libraries.
+
+What Can the APIs Do
+========================
+Essentially, each API is a language and hardware specific implementation of the IQUART protocol with access to information about each of the IQUART clients.
```
iq-fred, line 22: When you say "hardware specific", what do you mean? Is that referring specifically to the Arduino libraries? The Python, Matlab, and C++ ones are hardware agnostic I would say, and even the "arduino" one works across a pretty broad range of hardwares

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+.. _getting_started_python_api:
+
+******************************
+Getting Started with Python
+******************************
+
+Installation
+==================
+.. note::
+    We heavily recommend to manage your projects with some sort of virtual environment, for example Pipenv. For instructions to install Pipenv for Mac or Windows click `here <https://medium.com/@mahmudahsan/how-to-use-python-pipenv-in-mac-and-windows-1c6dc87b403e>`_. For installation instructions for Linux click `here <https://github.com/pypa/pipenv>`_.
```
iq-fred, line 13: This may be a bit outside of our purview to recommend the usage of virtual environments to people. May want to restrict ourselves to just things specific to our library.  So you may want to lose this note and the installing with Pipenv part below. 

I know it previously existed on our old Getting Started with Python, but honestly I don't see the point in us recommending virtual environments to people. We are just another library, let people manage their environments anyway they like. It makes no difference to us.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Installing with Pipenv
+-----------------------------
+In order to install the Vertiq python API into your pipenv virtual environment, simply run ``pipenv install iqmotion`` within your virtual environment.
+
+Installing with Pip
+-----------------------------
+In order to install the Vertiq python API via pip, simply run ``pip install iqmotion``.
```
iq-fred, line 21: Throwing down a link to our PyPi for more information for people who like to look at that stuff may be good here: https://pypi.org/project/iqmotion/

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+    By default, all modules have an ID of 0.
+
+3. Optionally, a firmware style (``firmware``)
+
+  * This value is a string, and can be “speed”, “servo”, or “pulsing”. To learn more about what firmware styles are available for your module, please refer to your module's family page.
+
+4. Optionally, a path to additional clients (``clients_path``)
+
+  * There are two options available for adding extra clients. First, in order to add all client files in a folder, you can pass a path to a folder that contains all additional, client json files. Second, 
+    you can pass an array of paths to custom client json files in order to only add those you are interested in rather than an entire folder.
```
iq-fred, line 64: It may be worth mentioning in a note or something that using additional clients is pretty uncommon for customers. I just don't want people to think they should be typically using additional client files.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+Creating a Module Object
+-----------------------------
+Vertiq's module objects are preconfigured to communicate with all accessible IQUART clients for your module's firmware. All module objects are created with the same 
+initialization.
+
+1. A serial communicator object
+2. Optionally, a module ID (``module_idn``)
+
+  * This represents the module ID of the device you would like to communicate with. If your module has its module ID configured to 26, you must set ``module_idn`` to 26 in object instantiation. 
+    By default, all modules have an ID of 0.
```
iq-fred, line 55: When you say "By default, all modules have an ID of 0", I'm a little unsure if you mean the module objects in the API have an ID of 0 by default, or if the actual module firmware has an ID of 0 by default. Obviously I know the answer is both, but here it may be worth clarifying a bit more, e.g. "If no module ID is specified, the module object is created with an ID of 0".

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+
+3. Optionally, a firmware style (``firmware``)
+
+  * This value is a string, and can be “speed”, “servo”, or “pulsing”. To learn more about what firmware styles are available for your module, please refer to your module's family page.
```
iq-fred, line 59: May be worth specifying what the default value of this is when nothing is specified. I assume its "speed", but I'm not really sure myself.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+
+.. code-block:: python
+
+    print(module.get_all("system_control"))
+
+Again, you can also treat the returned value as a normal parameter, and store it in a variable. In this case the data is stored as an array.
+
+Set
+^^^^^^^^^^
+All Python set commands have the format ``module.set("client", "client_entry", value)``. The set function changes the value of the target ``client_entry`` to ``value``.
```
iq-fred, line 142: This is actually not quite true. You can do module.set("client", "client_entry") with no value for some entries I believe, like module.set("system_control", "reboot_program"), and it will still do what you want. May be worth calling out the existence of that type of entry in your set section

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+^^^^^^^^^^^^
+All Python save commands have the format ``module.save("client", "client_entry")``. The save function takes the currently set entry value, and stores it in the module's persistent memory. 
+Values that are saved are retained on power cycles.
+
+Suppose we want to save the timeout value set above. To do so:
+
+.. code-block:: python
+
+    module.save("propeller_motor_control", "timeout")
+
```
iq-fred, line 165: There's also "set_verify" I believe, which does a set and then checks the value with a get. Not terribly important, but maybe worth shouting out its existence here.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,170 @@
...
+Next Steps
```
iq-fred, line 166: I know you link to an example over in the propeller motor control docs, but I just love coding examples. It may be worth adding a basic example here for the people who come to this page, read none of what you written, and just want to copy-paste a block of code to get started. Putting something very simple here, like something that just makes a communicator, makes a module, and gives an example of doing gets and sets. Maybe just set it to spin very slowly and sit in a while loop doing gets on speed or something. Or something else, anything that's a very basic easy demo of getting and setting, just something that a user on this page can plop into python and see that something is in fact happening before they move onto the propeller motor control stuff.

iq-fred (changes requested):

docs/api_docs/python.rst
```diff
@@ -0,0 +1,214 @@
...
+.. code-block:: python
+
+    module.set("system_control", "reboot_program")
+
+Set Verify
+^^^^^^^^^^^^
+Additionally to the standard set, you can use the ``set_verify`` function in order to set a new value, and confirm that the value has been set correctly on the module. Set verify 
+calls have the same format as standard set commands, but provide additional optional parameters ``get_values``, ``time_out``, ``retries``, and ``save``. 
+
+* ``get_values``: Specifies values to add in the get message (such as index for some client entries). By default, this is ``None``
```
vertiq-jordan, line 160: stole these from the API docstrings directly.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,214 @@
...
+.. image:: ../_static/api_pics/timeout_entry.png
+
+To do so:
+
+.. code-block:: python
+
+    module.set("propeller_motor_control", "timeout", 3)
+
+Some client entries (such as :ref:`System Control's <system_control>` ``reboot_program``) accept sets without a ``value``. Simply calling ``module.set("client", "client_entry")`` is enough to 
+trigger the desired behavior. So, to reboot your module's program:
```
iq-fred, line 149: May be worth specifying that you can tell which ones these are since their client table entries have no data type specified

docs/api_docs/python.rst
```diff
@@ -0,0 +1,214 @@
...
+
+Set Verify
+^^^^^^^^^^^^
+Additionally to the standard set, you can use the ``set_verify`` function in order to set a new value, and confirm that the value has been set correctly on the module. Set verify 
+calls have the same format as standard set commands, but provide additional optional parameters ``get_values``, ``time_out``, ``retries``, and ``save``. 
+
+* ``get_values``: Specifies values to add in the get message (such as index for some client entries). By default, this is ``None``
+* ``time_out``: A blocking timeout while verifying the set is successful in seconds. By default, this is 0.1s
+* ``retries``: The number of times you would like to retry the set before giving up. By default, this is 5
+* ``save``: Allows you to save the value once the set is confirmed. By default this is ``False``
```
iq-fred, line 163: An example of a set-verify would be nice for consistency

docs/api_docs/python.rst
```diff
@@ -0,0 +1,214 @@
...
+    time.sleep(5)
+
+    #Loop forever
+    while(1):
+
+        #Spin very slowly
+        module.set("propeller_motor_control", "ctrl_velocity", 20)
+
+        #Check our velocity
+        print(module.get("brushless_drive", "obs_velocity"))
```
iq-fred, line 211: A sleep here may be nice for being able to somewhat read the output. Something less than 0.1 seconds I guess to avoid timeout, maybe 50 ms or something.

docs/api_docs/python.rst
```diff
@@ -0,0 +1,214 @@
...
+.. code-block:: python
+
+    module.save("propeller_motor_control", "timeout")
+
+Next Steps
+=========================
+As the get, set, and save commands are the basis of all IQUART configuration and control, you now posess all of the base knowledge necessary to begin development 
+with the Vertiq Python API.
+
+A very basic example is provided here. It demonstrates the basics of setting up communication and a module object as well as how to ``set``, ``get``, and ``get_all``.
```
iq-fred, line 181: Maybe worth providing our standard, "remove props and secure motor disclaimer" here

iq-fred (changes requested):

iq-fred (approved):

PR #78, opened 2024-07-10
Feature/import resource pack

Vertiq-Ben:
- I updated the Adding Resource Files instructions using the Import Resource Pack feature and updated the related screenshots.
- There is a note at the beginning of the instructions to tell the user to use the 'sudo' command in order to run Control Center with administrator privileges on Mac and Linux.
- I also removed the old instructions for importing resource files from the Appendix

https://iqmotion.readthedocs.io/en/feature-import-resource-pack/control_center_docs/control_center_start_guide.html#adding-resource-files


docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -705,47 +705,38 @@ Once clicked, a File Explorer will open, and you can select a name and location
...
+3. Once the application runs, navigate to the Menu in the top left and select 'Import Resource Pack'.
 
-Linux (Ubuntu)
-------------------
-1. Open a new terminal window, and navigate to the IQ Control Center installation. By default this is Home/~user~/IQ Control Center
-2. Inside you will see the following
+.. image:: ../_static/control_center_pics/windows/windows_menu_import_resource_pack.png
 
-.. image:: ../_static/control_center_pics/linux/linux_install_folder.png
+4. A popup will appear requesting that you run Control Center as an administrator. Since you already did this in step 1, click 'OK' and proceed to the next step.
```
vertiq-jordan, line 728: As a future control center idea...is there an easy way for us to tell if the app has already been open as an admin so we can avoid this popup if it's unnecessary?

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -705,47 +705,38 @@ Once clicked, a File Explorer will open, and you can select a name and location
...
+.. image:: ../_static/control_center_pics/windows/windows_resource_pack_file_explorer.png
 
-.. image:: ../_static/control_center_pics/mac/mac_in_the_install.png
+6. After the .zip file is imported, a popup will appear requesting that you restart Control Center. Click 'OK' and restart Control Center, but this time you do not have to run as an administrator.
 
-3. Enter IQ Control Center.app/Contents/MacOS. Inside you will see the following
+.. image:: ../_static/control_center_pics/windows/windows_application_restart_required_popup.png
 
-.. image:: ../_static/control_center_pics/mac/mac_macos_folder.png
+7. If the provided resource .zip file includes default files, you will find them in the dropdown menu in the General tab next to MODULE PRE-SETTINGS. The other resource files will include additional configuration options based on your connected module.
```
vertiq-jordan, line 740: Could we link "defaults files" up to the Configuration with Defaults section above just in case people don't know what "default files" means?

docs/control_center_docs/control_center_start_guide.rst
vertiq-jordan, line 728: ...requesting that you run the Control Center...

Vertiq-Ben: Currently it does not show any indication that the app is running with admin privileges but this is a good idea. One way would be to try running a command that requires admin privileges and display something on the application depending on if it is successful or not.

vertiq-jordan (approved):

docs/manual/manual_appendix.rst
```diff
@@ -30,54 +30,6 @@ For example, if the stopping time were 1 second and the stopping speed were 5 ra
 and then 2 rad/s again for another 0.5 seconds, it would not be considered stopped. The countdown was reset when the module went to 10 rad/s. If the module 
 stayed at 2 rad/s for 1 second, it would then be considered stopped.
```
Vertiq-Ben, line 32: Removed old instructions below

docs/control_center_docs/control_center_start_guide.rst
```diff
@@ -705,47 +705,36 @@ Once clicked, a File Explorer will open, and you can select a name and location
...
+.. note::
+    The following instructions apply to all supported operating systems: Windows, Mac, and Linux (Ubunutu). The first step requires the Control Center application to be opened with administrator privileges.
+    On Mac and Linux, you must use the 'sudo' command in the terminal to run the application.
 
-.. image:: ../_static/control_center_pics/windows/windows_installed_folder.png
+    .. image:: ../_static/control_center_pics/linux/linux_sudo_command.png
 
-3. Enter the IQ Control Center folder, and you’ll see the Resources folder
+1. Use the File Explorer to locate the Control Center's installation. By default, the application is located in Program Files (x86)/IQ Control Center/IQ Control Center/IQ Control Center.exe
+2. Right click the executable and select 'Run as administrator'.
```
iq-fred, line 718: @Vertiq-Ben Step 1 and 2 here are Windows specific I believe. Whereas the stuff for launching on Mac or Linux are in a note above these instructions. These numbered steps are not platform agnostic. Maybe it would be better to have a step that is "Run the Control Center as Administrator", and then as subpoints to that say you could have the instructions for Mac and Linux, and the instructions for Windows

iq-fred (changes requested):

iq-fred (approved):

PR #81, opened 2024-07-17
Add Safety Manual

vertiq-jordan:
built docs here: https://iqmotion.readthedocs.io/en/feature-add_safety_docs/manual/manual_safety_systems.html


docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+Suppose you set your otw to 140°C and your otlo to 150°C. This means that there will be no derate limiting applied under 140°C, your module will stop spinning at and above 150°C, and will be derate limited in between.
+
+A graph illustrating the microcontroller temperature derate calculation is provided below:
+
+.. image:: ../_static/manual_images/safety/coil_temp_derate.png
+    :scale: 50%
+
+Calculation of the Total Derate
+-----------------------------------
+The total applied derate applied is the direct multiplication of the speed, microcontroller, and coil derate values. So, the our drive voltage system becomes:
```
iq-fred, line 91: Has "the" instead of what was probably meant to be "then" in the second sentence: "So, the our drive voltage..."

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+.. image:: ../_static/manual_images/safety/full_derate_system.png
+    :scale: 50%
+
+Power Safety
+=================
```
iq-fred, line 97: Might be worth mentioning as a first thing here sort of in general what the point of power safety is. As in, its a last-line of defense emergency safety that monitors certain parameters of the motor and will immediately cause it to coast if any of them reach a dangerous level. Maybe not in those exact words, but some basic background on what this is and why it exists outside of the normal derate stuff could be good as an intro.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+The full listing of power safety parameters can be found in the :ref:`Power Safety Client table <power_safety>`. Each safety has a low end limit as well as a high end limit. 
+If the measured parameter drops below the low end or goes above the high end, a fault flag is raised, and can be viewed in the ``fault_now`` parameter. 
+The ``fault_now`` parameter is only set by hardware. All faults that have occurred since the module's start up can be viewed in the ``fault_ever`` parameter. 
+The value of ``fault_ever`` is generally set by the hardware and cleared by the user.
+
+The module's behavior when faults are detected is determined by the ``fault_latching`` parameter. If ``fault_latching`` is set to 1, then any high flags 
+in ``fault_ever`` place the module into safe mode. The module remains in safe mode until the user clears ``fault_ever``. When ``fault_latching`` is set to 0, 
+the module remains in safe mode until all power safety parameters are back in their safe operating range, and ``fault_now`` returns to 0.
```
iq-fred, line 110: May be worth clarifying somewhere what exactly "safe mode" is from a user perspective. Basically it coasts the motor and it won't spin or apply any voltage while its in safe mode. Just want to make it clear to users what exactly being in safe mode implies

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+Internal Programmable Voltage Detector (PVD)
+------------------------------------------------
+Each module's microcontroller has an internal PVD that monitors the microcontroller's Vdd voltage level. In the event that Vdd drops below the pre-programmed threshold 
+value (~2.8V), the module is forced to stop spinning. The module will only be allowed to resume normal operation when Vdd has risen back above the voltage threshold (~2.9V).
```
iq-fred, line 115: From a customer perspective, assuming they aren't an electrical or microcontroller person, what does Vdd voltage level mean? Maybe good to define it for the layman and give some context, like "the voltage that is used to power the microcontroller. If this drops too low the microcontroller on your module cannot function properly".

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+The full listing of power safety parameters can be found in the :ref:`Power Safety Client table <power_safety>`. Each safety has a low end limit as well as a high end limit. 
+If the measured parameter drops below the low end or goes above the high end, a fault flag is raised, and can be viewed in the ``fault_now`` parameter. 
+The ``fault_now`` parameter is only set by hardware. All faults that have occurred since the module's start up can be viewed in the ``fault_ever`` parameter. 
+The value of ``fault_ever`` is generally set by the hardware and cleared by the user.
+
+The module's behavior when faults are detected is determined by the ``fault_latching`` parameter. If ``fault_latching`` is set to 1, then any high flags 
+in ``fault_ever`` place the module into safe mode. The module remains in safe mode until the user clears ``fault_ever``. When ``fault_latching`` is set to 0, 
+the module remains in safe mode until all power safety parameters are back in their safe operating range, and ``fault_now`` returns to 0.
+
```
iq-fred, line 111: I'm a sucker for examples, any chance we could describe a power safety one here? Like "Suppose the high end limit for the microcontroller temperature power safety parameter is set to 115C. If at your module's microcontroller temperature is ever at or above 115C, the module will enter safe mode and will not spin until it exits safe mode. If fault latching is set blah blah blah, if fault latching is not set blah blah blah" you get the idea

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+Internal Programmable Voltage Detector (PVD)
+------------------------------------------------
+Each module's microcontroller has an internal PVD that monitors the microcontroller's Vdd voltage level. In the event that Vdd drops below the pre-programmed threshold 
+value (~2.8V), the module is forced to stop spinning. The module will only be allowed to resume normal operation when Vdd has risen back above the voltage threshold (~2.9V).
+
+.. _slew_rate:
+
+Slew Rate
```
iq-fred, line 119: As a general terminology thing in this section, I would say that "slew rate" itself is not the thing that does the limiting, slew rate just means rate of change. I would probably call this the "Slew Rate Limiter" and a "slew rate limit" throughout here. Pedantic, I know, but I just like to keep the distinction between the concept of a slew rate and something that limits that slew rate

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+===========
+Your module's slew rate defines the maximum change in control voltage per second. Slew rate works as a form of software filter that ensures that your module does not
```
iq-fred, line 121: I won't put it everywhere, but to illustrate what I mean with the comment above, this should probably be "Your module's slew rate limit..."

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+------------------------------------------------
+Each module's microcontroller has an internal PVD that monitors the microcontroller's Vdd voltage level. In the event that Vdd drops below the pre-programmed threshold 
+value (~2.8V), the module is forced to stop spinning. The module will only be allowed to resume normal operation when Vdd has risen back above the voltage threshold (~2.9V).
+
+.. _slew_rate:
+
+Slew Rate
+===========
+Your module's slew rate defines the maximum change in control voltage per second. Slew rate works as a form of software filter that ensures that your module does not 
+attempt to change speeds faster than is safe.
```
iq-fred, line 122: Hmm, I may change up the second sentence a bit. Usually the slew rate is their mainly to protect other electronics or flight controllers, not the module's themselves. They are generally fine with no slew limit set, so I don't want to give that impression. Maybe more like "The slew rate limiter can be used to limit how quickly your module attempts to change speed, which can be useful to protect external electronics or improve integration with flight controller's in certain cases."

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+.. image:: ../_static/manual_images/safety/slew_rate_1.png
+    :scale: 80%
+
+As you can see, the slew limit parameter helps protect against sudden changes in commanded voltage. This can be especially useful in :ref:`noisy/vibration heavy 
+environments <vibration_and_jittering>`. The amount of slew filtering that is necessary (if any) is completely application specific, and will require manual tuning.
+
+Brushless Drive Current Limiter
+====================================
+:ref:`Brushless Drive's <brushless_drive>` ``motor_I_max`` defines that maximum allowable drive current. In general, this current limit protects your module against extreme step commands, and ensures that the current moving through the motor is always in a safe region.
```
iq-fred, line 161: Should this be "the maximum allowable..." instead of "that maximum allowable..."?

iq-fred, line 161: A couple things content-wise on this first bit. You do say it and I know that you know it, but it may be worth driving home because people often get confused on this: this is the maximum allowable motor current, not supply current. They are not necessarily the same.

About the second sentence, it does protect us against extreme step commands, and specifically it's meant to protect the module's electronics from current spikes that may permanently damage them. It's also useful for protecting them during stalls, which don't really involve an actual step command, but do lead to a huge surge in current. So maybe just don't frame it as just for steps, though they are part of it.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+This behavior is illustrated in the examples below. The first is the module with a ``motor_I_max`` of 50A. It is given a step to 5V, then after 5 seconds is stepped back to a 0V control.
```
iq-fred, line 162: Should this be "a module with a..." instead of "the module with a..."?. Also the "stepped back to a 0V control" sounds a bit odd, maybe "stepped back down to a 0V command"?

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+
+.. image:: ../_static/manual_images/safety/current_limit_50.png
+    :scale: 80%
+
+Now, we set the current limit to 0.5A. You will see that the motor's rate of change is much slower. As mentioned above, this is meant to decrease the likelihood of dangerous currents as the module steps between velocities.
+
+.. image:: ../_static/manual_images/safety/current_limit_0_5.png
+    :scale: 80%
+
+The biggest difference between this current limit and :ref:`slew_rate` is that the slew rate parameter defines an absoulte linear relation between the control voltage applied and
```
iq-fred, line 172: "absoulte" instead of "absolute"

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+    :scale: 80%
+
+Now, we set the current limit to 0.5A. You will see that the motor's rate of change is much slower. As mentioned above, this is meant to decrease the likelihood of dangerous currents as the module steps between velocities.
+
+.. image:: ../_static/manual_images/safety/current_limit_0_5.png
+    :scale: 80%
+
+The biggest difference between this current limit and :ref:`slew_rate` is that the slew rate parameter defines an absoulte linear relation between the control voltage applied and 
+the time it takes to get there. This current limit is not meant to limit the rate of change of commanded voltages, it is meant to protect against sudden voltage 
+spikes due to large voltage steps.
```
iq-fred, line 174: There's a bit more to it than this, as I touched on in my comment above. The current max limit does a little more for us than just controlling steps like the slew limit, it more generally protects us from high current events, like a stall. For a colorful example, I once threw a towel at a spinning motor to test its I Max. We didn't change its command or drive voltage at all, it just couldn't spin, and so no more back-EMF and so loads of current. Maybe that's what you were going for here as well when you say "voltage steps", but it may be good to clarify that difference of it a bit more. I'm happy to sit down and talk about this more in person if you want some more insight on what the differences are

iq-fred (changes requested):

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+*****************
+Safety Systems
+*****************
+The following manual is meant to outline the safety systems present on Vertiq modules. Safety systems are built into your module's firmware, 
+and seek to protect your module from things like velocity runaway, power issues, and overheating.
+
+Protecting Against Voltage Regeneration Spikes
+====================================================
+Voltage regeneration is one of the most often cited reasons for module failure, and as such we have advanced systems to protect against extreme voltage spikes. 
+You can read more about voltage regeneration protection in :ref:`protect_against_regen`.
```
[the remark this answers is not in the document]
vertiq-jordan: I'm fine with moving it into here. It's not really a tutorial anyway, so it probably makes more sense in here.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,174 @@
...
+brushless drive's ``motor_redline_start`` parameter, and brushless drive's ``motor_redline_end`` parameter. You can find the redline parameters listed in the 
+:ref:`Brushless Drive Message Table <brushless_drive>`.
+
+The total speed derate is calculated by :math:`\text{Speed Derate} = \frac{\text{motor_redline_end - speed}}{\text{motor_redline_end - motor_redline_start}}` where the solution is saturated to [0,1].
+
+This means that for any speed below the range [redline_start, redline_end], there is no derate applied. For any speed above the range [redline_start, redline_end], 
+the speed derate is 0. Any speed in that range is subject to a variable derate [0, 1].
+
+Suppose you set your redline start to 1750 rad/s and your redline end to 2000 rad/s. This means that there will be no derate limiting applied under 1750 rad/s, 
+your module will stop spinning at and above 2000 rad/s, and will be derate limited in between.
```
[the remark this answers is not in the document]
vertiq-jordan: We really ought to go through the control center and pull out things that "customer's shouldn't generally be using." If we really really don't want them touching things, we shouldn't give a big button to change it (even if it's behind the very scary advanced tab). If someone comes to us with an issue, and we determine it can be fixed by changing the redlines, then we can just tell them how to do it with the API.

docs/tutorials/protect_against_regen.rst
vertiq-jordan, line 1: This moved into the general safety manual.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,255 @@
...
+    :scale: 50%
+
+There are three separate derate values that are internally updated, each linked with a different measured module parameter (e.g. speed and temperature). There is a speed derate used to avoid mechanical failure, 
+a microcontroller temperature based derate to avoid microcontroller overheating, and a coil temperature derate to avoid coil overheating. 
+These three derate values are combined and applied to the target command voltage in order to protect your module. Details of each of these derate values as well as 
+how they are combined are discussed below.
+
+Speed Based Derate
+---------------------
+As mentioned above, your module's speed derate is meant to protect your module against speed runaway. Speed derate is calculated using the module's current speed,
```
iq-fred, line 81: Going along with the change above from last time, this shouldn't say its to protect against speed runaway anymore.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,255 @@
...
+
+A graph illustrating the microcontroller temperature derate calculation is provided below:
+
+.. image:: ../_static/manual_images/safety/coil_temp_derate.png
+    :align: center
+    :scale: 50%
+
+Calculation of the Total Derate
+-----------------------------------
+The total applied derate applied is the direct multiplication of the speed, microcontroller, and coil derate values. So, then our drive voltage system becomes:
```
iq-fred, line 151: This has applied twice in the first bit, probably only needs it once, the part I mean is "The total applied derate applied is...."

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Slew Rate Limiter
+==================
+Your module's slew rate limit defines the maximum change in control voltage per second. The slew rate limiter can be used to limit how quickly your module attempts to 
+change speed, which can be useful to protect external electronics or improve integration with flight controllers in certain cases.
+
+There are two parameters in the :ref:`Brushless Drive client <brushless_drive>` that affect your module's slew rate limiting:
+
+* ``slew_volts_per_second`` defines the actual slew rate in volts per second applied to voltage commands. This defines the strength of the filter's impact.
+* ``slew_enable`` determines whether or not ``slew_volts_per_second`` is actually applied to voltage commands. If ``slew_enable`` is false, then there will not be slew limiting applied.
```
iq-fred, line 196: I think these slew guys are also available through the Control Center, may be worth showing that they can be configured through there as well, just showing that it exists at least

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. image:: ../_static/manual_images/safety/current_limit_0_5.png
+    :align: center
+    :scale: 80%
+
+The biggest difference between this current limit and :ref:`slew_rate` is that the slew rate parameter defines an absolute linear relation between the control voltage applied and 
+the time it takes to get there. This current limit is not meant to limit the rate of change of commanded voltages, it is meant to protect against all high current events. Suppose 
+your module is spinning, but is caught in a tree and can no longer spin. Assuming that your flight controller continues to send commands, your module will continue to 
+attempt spinning, but if it cannot, then it will end up with a surge of current going through its coils. In this case, ``motor_I_max`` would protect the module by 
+forcing it to stop trying to spin.
```
iq-fred, line 255: It actually won't make it not try to spin necessarily, it will just limit the amount of current that is being asked for by limiting the drive voltage. If you pulled it off that tree, it would start spinning again once nothing was blocking it.

iq-fred (changes requested):

iq-fred (approved):

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,264 @@
...
+.. _protect_against_regen:
+
+Protecting Against Voltage Regeneration Spikes
+====================================================
+What are Regenerative Voltage Spikes
+------------------------------------------------------------
+Regeneration voltage occurs when the commanded module voltage is less than the module's back-EMF voltage. When this occurs, the motor's kinetic 
+energy is converted directly to electrical energy, and produces a current flowing out of the motor and into the power source.
+
+The commanded module voltage may be less than the back-EMF voltage in cases where the module is slowing down or braking. The faster that the commanded voltage
```
iq-matt, line 21: It's not a rate relationship, it's a difference relationship.  "A larger spread between back-EMF and commanded module voltage leads to a greater magnitude of regeneration current".

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,264 @@
...
+
+A graph illustrating the speed derate calculation is provided below:
+
+.. image:: ../_static/manual_images/safety/speed_derate_graph.png
+    :align: center
+    :scale: 50%
+
+Microcontroller Temperature Based Derate
+--------------------------------------------
+As mentioned above, your module's microcontroller temperature derate is meant to protect your module's microcontroller from overheating.
```
iq-matt, line 114: We should have a warning in this section that notes that this temperature derate only works when motoring.  Successful thermal derating is not guaranteed when generating.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,264 @@
...
+
+A graph illustrating the microcontroller temperature derate calculation is provided below:
+
+.. image:: ../_static/manual_images/safety/micro_temp_derate_graph.png
+    :align: center
+    :scale: 50%
+
+Coil Temperature Based Derate
+---------------------------------
+As mentioned above, your module's coil temperature derate is meant to protect your module's coils from overheating. Coil temperature derate is calculated using the
```
iq-matt, line 134: This section needs the same warning about how it only works when motoring.

docs/manual/manual_safety_systems.rst
```diff
@@ -0,0 +1,264 @@
...
+
+The full listing of power safety parameters can be found in the :ref:`Power Safety Client table <power_safety>`. Each safety has a low end limit as well as a high end limit. 
+If the measured parameter drops below the low end or goes above the high end, a fault flag is raised, and can be viewed in the ``fault_now`` parameter. 
+The ``fault_now`` parameter is only set by hardware. All faults that have occurred since the module's start up can be viewed in the ``fault_ever`` parameter. 
+The value of ``fault_ever`` is generally set by the hardware and cleared by the user.
+
+The module's behavior when faults are detected is determined by the ``fault_latching`` parameter. If ``fault_latching`` is set to 1, then any high flags 
+in ``fault_ever`` place the module into safe mode. In safe mode, the module is set to coast, and it will not spin or apply any control voltage. The module remains in safe mode until the user clears ``fault_ever``. 
+When ``fault_latching`` is set to 0, the module remains in safe mode until all power safety parameters are back in their safe operating range, and ``fault_now`` returns to 0.
```
iq-matt, line 175: When fault_latching is set to 0, the module can come in and out of safe mode at high frequency.  This may be damaging to the the motor or supply since it can cause large ripple currents.  Vertiq recommends using fault_latching 1 whenever possible.

iq-matt (changes requested):

iq-matt (approved):

PR #87, opened 2024-07-31
Add DroneCAN Tunnel Docs

vertiq-jordan:
Adds documentation about the DroneCAN tunnel message to the DroneCAN protocol page.


docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -117,6 +117,141 @@ The values represent the speed and direction to command the motor to, normalized
 Refer to the `uavcan.equipment.esc.RawCommand section of Standard Data Types <https://dronecan.github.io/Specification/7._List_of_standard_data_types/#rawcommand>`_ in the specification 
 for more details
 
+uavcan.tunnel.Broadcast (Data Type ID = 2010)
```
vertiq-jordan, line 120: I chose to add this to where we already talk about all of the supported messages. This one is slightly unique because it has an example with it

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -117,6 +117,141 @@ The values represent the speed and direction to command the motor to, normalized
...
+
+.. image:: ../_static/manual_images/dronecan/dronecan_tunnel_python_sequence.png
+
+The Python script necessary to accomplish this is provided here: 
+
+.. code-block:: python
+
+	import time
+	import iqmotion as iq
+	import pyuavcan_v0 as uavcan
```
iq-fred, line 146: Any chance you could see if we can use pydronecan instead? This is an old library, and I feel like we will get laughed at by the all cool dronecan kids if they see it: https://github.com/dronecan/pydronecan

iq-fred (changes requested):

iq-fred (approved):

PR #94, opened 2024-08-15
Bugfix/typos and v max start

Vertiq-Ben:
This PR includes many bugfixes:

- remove save from fault_ever in power safety
- added v_max_start in brushless_drive
- fixed notes field of stow_status
- fixed cpp full examples reference


docs/clients/brushless_drive.rst
```diff
@@ -168,6 +168,8 @@ Type ID 50 | Brushless Drive
...
         +--------+-------------------------------+----------------+-----------+----------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+        | 29     | v_max_start                   | get, set, save | float     | :math:`V`                              | This is the voltage at which the safeties will start to attempt to limit the regeneration of the motor.                                                                                                                                                                                                                                                                                                     |
```
iq-fred, line 171: I think it would be good to mention this works in conjunction with volts_limit, and that its in terms of supply voltage. Maybe something like:

"This is the supply voltage at which the safeties will start to attempt to limit the regeneration current from the motor. This is used in conjunction with the volts_limit parameter, v_max_start sets the starting voltage for the safety, and volts_limit sets the ending voltage."

May be good to add that working in conjunction bit to the volts_limit parameters description as well while we're in here

iq-fred (changes requested):

iq-fred (approved):

PR #122, opened 2025-01-09
Add redundant throttle docs

vertiq-jordan:
What the title says...built docs here: https://iqmotion.readthedocs.io/en/feature-redundant_throttle/manual/manual_redundant_throttle.html

I'm targeting the "add new dronecan docs" branch since redundant throttle and those features will go out at the same time, and it's already a holding branch.


docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Vertiq's supported throttle sources are :ref:`DroneCAN <dronecan_protocol>`, :ref:`IQUART <uart_messaging>`, and :ref:`hobby protocols <hobby_protocol>`. Please note that only priority values corresponding to protocols supported by 
+your module have any effect. See your module's family page to see what throttle sources are supported.
+
+Priorities can be valued from 0 to 3. Setting a priority of 0 indicates that the module will ignore all throttle messages received from the configured source. 
+Priority values [1, 3] define the priority of each protocol against the others where 3 defines the highest priority.
+
+.. note:: 
+    Due to Vertiq modules' hardware, it is not possible to use both IQUART and hobby protocols simultaneously. As such, it is only possible to use DroneCAN with one of IQUART 
+    or hobby as redundant throttle sources.
```
Vertiq-Ben, line 47: Does this mean that DroneCAN must be the main throttle source, and either IQUART or hobby is the secondary?

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,112 @@
...
+
+- IQUART's priority to 3 and DroneCAN's to 1 (the module will automatically set hobby's to 2 ensuring that there are no two matching priorities)
+- Flight controller outputting both :ref:`IQUART Flight Controller Interface commands <controlling_ifci>` as well as DroneCAN commands
+- Throttle source timeout configured to 0.5 seconds
+
+The sequence diagram below illustrates how your module reacts to received throttle commands as well as how it deals with switching between sources.
+
+.. image:: ../_static/manual_images/redundant_throttle/iquart_prio_example.png
+    :width: 50%
+
```
Vertiq-Ben, line 64: When IQUART disconnects, the flight controller still sends IQUART commands right? It's just that none of the commands will reach the module due to the severed connection. Is performance hindered?

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,112 @@
...
+- DroneCAN's priority to 3, IQUART to 0, hobby to 0
+- Flight controller outputting both :ref:`hobby_dshot` as well as DroneCAN commands
+- Throttle source timeout configured to 0.1s
+
+.. image:: ../_static/manual_images/redundant_throttle/dshot_ignored_ex.png
+    :width: 70%
+
+In this example, the module is configured to ignore all hobby and IQUART throttle messages. So, all received DSHOT throttles are dropped, and throttles are only applied once DroneCAN 
+throttles are received. An important note from this example is that the module will reach the module's :ref:`propeller motor control timeout <manual_timeout>` even if the
```
Vertiq-Ben, line 83: Maybe rephrase this sentence to: "An important note from this example is that the module will reach **its** propeller motor control timeout even if the flight controller continues to send DSHOT throttles.

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,112 @@
...
+- Flight controller outputting both :ref:`IQUART Flight Controller Interface commands <controlling_ifci>` as well as DroneCAN commands
+- Throttle source timeout configured to 0.5 seconds
+
+The sequence diagram below illustrates how your module reacts to received throttle commands as well as how it deals with switching between sources.
+
+.. image:: ../_static/manual_images/redundant_throttle/iquart_prio_example.png
+    :width: 50%
+
+Note that the first DroneCAN throttle is applied to the motor since no IQUART messages had been received before. Then, when both an IQUART and DroneCAN message are received, 
+the IQUART throttle is applied. Last, DroneCAN throttles are only applied again once the throttle timeout is reached after IQUART disconnects.
```
Vertiq-Ben, line 66: Is it possible for IQUART to reconnect for whatever reason? If so, what happens?

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    <style type="text/css">
+    .center_vid {   margin-left: auto;
+                    margin-right: auto;
+                    display: block;
+                    width: 75%; 
+                }
+    </style>
+    <video class='center_vid' controls><source src="../_static/manual_images/redundant_throttle/redundant_arming_example.mp4" type="video/mp4"></video>
+
```
Vertiq-Ben, line 111: No issue here, this is a really cool example!

docs/clients/throttle_source_manager.rst
```diff
@@ -0,0 +1,123 @@
...
+        +========+========================+================+===========+=================================+===================================================================================================================================================================================+
+        | 0      | throttle_timeout       | get, set, save | float     | Seconds                         | Defines the amount of time, in seconds, that the Throttle Source Manager will wait for a new message from the highest received priority before checking the next lowest priority. |
+        +--------+------------------------+----------------+-----------+---------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+        | 1      | dronecan_priority      | get, set, save | uint8     |                                 | The relative priority of DroneCAN throttles against hobby and IQUART throttles. Must be a value [0, 3] where 0 means DroneCAN is ignored, and 3 is the highest priority.          |
```
Vertiq-Ben, line 117: Should the Unit for the priority values be Enum?

vertiq-jordan: Nope, the priorities are just numbers, they don't map to any special treatment

docs/manual/manual_redundant_throttle.rst
vertiq-jordan: yes and no. Depending on your priority settings, what you said is true, but it could also be that IQUART or hobby is primary and DroneCAN is secondary

docs/manual/manual_redundant_throttle.rst
vertiq-jordan: It depends on the failure honestly. If the wire just gets cut, then the flight controller will keep sending UART messages...that just won't go anywhere. If there's a software failure or the flight controller's serial peripheral explodes, it could just stop sending IQUART altogether. If your throttle timeout is set to something reasonable, there is no performance impact in this case. The flight controller will output the same throttle value on both IFCI and DroneCAN, so when it switches over to DroneCAN, it'll keep the same command it had from IQUART

Vertiq-Ben (approved):

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
+##############################
+Redundant Throttle Support
+##############################
+
+***************************
+About Redundant Throttle
+***************************
+
+Vertiq's redundant throttle support is a safety feature that allows your module to accept multiple throttle sources' commands simultaneously, and to switch between them should
```
iq-fred, line 9: When you say "throttle sources'" here, you may want to link over to where we explain what a throttle source is (https://iqmotion.readthedocs.io/en/feature-redundant_throttle/manual/manual_throttle.html#throttle-sources)

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+reacts to new throttle commands, and the module will eventually :ref:`manual_timeout`.
+
+Now, suppose that you connect DroneCAN as before, but now, also connect your module to a DSHOT output on your flight controller.
+
+.. image:: ../_static/manual_images/redundant_throttle/dshot_and_dronecan_connection.png
+    :width: 50%
+
+Under normal operating conditions, the module will apply only throttle commands received via DroneCAN (as determined by your redundant throttle configuration discussed :ref:`below <redundant_throttle_config>`). 
+If we once again sever the DroneCAN connection, however, the module will seamlessly transition to accepting DSHOT commands. By connecting multiple throttle sources, and 
+leveraging Vertiq's redundant throttle feature, you robustify your vehicle's ability to fly safely even after communication failures.
```
iq-fred, line 26: On the one hand, it turns out robustify is actually a word from my googling, but on the other hand, it still feels weird to me. Maybe "you make your vehicle more robust against communication failures in flight" could work instead?

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Vertiq's supported throttle sources are :ref:`DroneCAN <dronecan_protocol>`, :ref:`IQUART <uart_messaging>`, and :ref:`hobby protocols <hobby_protocol>`. Please note that only priority values corresponding to protocols supported by 
+your module have any effect. See your module's family page to see what throttle sources are supported.
+
+Priorities can be valued from 0 to 3. Setting a priority of 0 indicates that the module will ignore all throttle messages received from the configured source. 
+Priority values [1, 3] define the priority of each protocol against the others where 3 defines the highest priority.
+
+.. note:: 
+    Due to Vertiq modules' hardware, it is not possible to use both IQUART and hobby protocols simultaneously. As such, it is only possible to use DroneCAN with one of IQUART 
+    or hobby as redundant throttle sources.
```
iq-fred, line 47: I may just say "it is not currently possible", since one day maybe we'll fix this.

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+
+
+.. _redundant_throttle_config:
+
+***************************************
+Redundant Throttle Configurations
+***************************************
+
+There are two types of configurations available with the redundant throttle. The first defines the amount of time that the module waits before switching to another throttle 
+source if the active source goes offline. The second set of parameters defines the priorities of each supported throttle source. You can find the specific parameters
```
iq-fred, line 36: What does "active source" mean? I know you and I know, but may be good to define that term a bit for people who are seeing this for the first time.

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+
+.. _redundant_throttle_config:
+
+***************************************
+Redundant Throttle Configurations
+***************************************
+
+There are two types of configurations available with the redundant throttle. The first defines the amount of time that the module waits before switching to another throttle 
+source if the active source goes offline. The second set of parameters defines the priorities of each supported throttle source. You can find the specific parameters 
+at :ref:`throttle_source_manager`.
```
iq-fred, line 37: If its possible to set these through Control Center, I always like to show it just to remind people that they can do it and not have to go through the API. So I guess a pitch for shoving in a mention of Control Center and a picture of these in there

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+your module have any effect. See your module's family page to see what throttle sources are supported.
+
+Priorities can be valued from 0 to 3. Setting a priority of 0 indicates that the module will ignore all throttle messages received from the configured source. 
+Priority values [1, 3] define the priority of each protocol against the others where 3 defines the highest priority.
+
+.. note:: 
+    Due to Vertiq modules' hardware, it is not possible to use both IQUART and hobby protocols simultaneously. As such, it is only possible to use DroneCAN with one of IQUART 
+    or hobby as redundant throttle sources.
+
+Configuration Example 1
```
iq-fred, line 49: Can I make a pitch for either adding a configuration example or adapting one of these to be one where its successfully using DroneCAN and hobby, in the most straightforward possible? I say this because I imagine the most common use case will be using DroneCAN and hobby, and I want an example of that to point to. So one where you put DroneCAN at priority 3, and hobby at 2, and whatever we recommend as a good timeout. So when someone asks us what we recommend for running DroneCAN and DSHOT together we can say "Do this example". Maybe that could be the new example 1 and you could shift your existing examples down to start with a super common case? Then the next example (your current example 1 as is) gets IQUART in there and shows how you can skip over priorities, and last one shows how you can use the ignore me priorities for a nice full set

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+throttles are received. An important note from this example is that the module will reach its :ref:`propeller motor control timeout <manual_timeout>` even if the 
+flight controller continues to send DSHOT throttles. As they are not processed by the throttle source manager, DSHOT throttles cannot be used to reset the timeout timer.
+
+********************************
+Redundant Throttle and Arming
+********************************
+
+All throttle commands received by the throttle source manager are sent to your module's :ref:`arming handler <manual_advanced_arming>`, and are subject to all constraints defined by the arming handler. 
+For example, suppose you have configured IQUART to be the highest priority source, and are sending IQUART throttle commands via the :ref:`Vertiq Testing Tool <vertiq_testing__guide>`. At the same 
+time, you are transmitting DroneCAN throttles with the :ref:`DroneCAN GUI tool <dronecan_gui_basics>`. In this example, your arming handler is set to arm on throttle in the range 0-7.5% on 10
```
iq-fred, line 93: the "set to arm on throttle in the range...." may be better as "set to arm on throttles". The parameter is called arm on throttle, but in this sentence we are arming on throttles in the range 0 to 7.5% I think

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+In this example, the module is configured to ignore all hobby and IQUART throttle messages. So, all received DSHOT throttles are dropped, and throttles are only applied once DroneCAN 
+throttles are received. An important note from this example is that the module will reach its :ref:`propeller motor control timeout <manual_timeout>` even if the 
+flight controller continues to send DSHOT throttles. As they are not processed by the throttle source manager, DSHOT throttles cannot be used to reset the timeout timer.
+
+********************************
+Redundant Throttle and Arming
+********************************
+
+All throttle commands received by the throttle source manager are sent to your module's :ref:`arming handler <manual_advanced_arming>`, and are subject to all constraints defined by the arming handler. 
+For example, suppose you have configured IQUART to be the highest priority source, and are sending IQUART throttle commands via the :ref:`Vertiq Testing Tool <vertiq_testing__guide>`. At the same
```
iq-fred, line 92: Should add a (VTT) after Vertiq Testing Tool, to define the acronym if you are going to use that acronym later on

docs/manual/manual_redundant_throttle.rst
iq-fred, line 93: May be better to link to the general external DroneCAN GUI Tool docs (https://dronecan.github.io/GUI_Tool/Overview/) for the DroneCAN GUI Tool link. The thing you link to now does give some basics on it, but then is really focused on firmware upgrading, which I feel like may be a bit weird for people to get sent over to a firmware upgrade page.

docs/manual/manual_redundant_throttle.rst
```diff
@@ -0,0 +1,113 @@
...
+********************************
+
+All throttle commands received by the throttle source manager are sent to your module's :ref:`arming handler <manual_advanced_arming>`, and are subject to all constraints defined by the arming handler. 
+For example, suppose you have configured IQUART to be the highest priority source, and are sending IQUART throttle commands via the :ref:`Vertiq Testing Tool <vertiq_testing__guide>`. At the same 
+time, you are transmitting DroneCAN throttles with the :ref:`DroneCAN GUI tool <dronecan_gui_basics>`. In this example, your arming handler is set to arm on throttle in the range 0-7.5% on 10 
+consecutive throttles.
+
+VTT outputs IQUART throttles at 10%, and the DroneCAN GUI transmits 3% throttles. In this case, your module will not arm. Since IQUART has a higher priority, its throttles take 
+precedence over DroneCAN, so the arming handler will only receive the 10% throttles. If you pause the VTT output, however, your module will arm as the throttle source manager will 
+have transitioned to accepting the DroneCAN commands within the arming region. Now, if you unpause the VTT output, your module will start spinning at 10% throttle.
```
iq-fred, line 98: I may add a bit more explanation at the end here to clarify a bit further for people who aren't as deep in the weeds as us. Something like:
"...within the arming region. Now that the module is armed, it will remain armed, even if the throttle source switches. If you unpause the VTT output your module will start spinning at 10% throttle because it is armed and receiving 10% throttle commands over IQUART, which have a higher priority than the DroneCAN throttle commands.

iq-fred (changes requested):

docs/manual/manual_redundant_throttle.rst
vertiq-jordan: ha...I did the same google search after I wrote it...I'll change it

iq-fred:
@vertiq-jordan General question, is any of the stuff on the existing arming pages outdated or inaccurate now? I recall that redundant throttle changes up some of the arming stuff. Do we need to update any of those pages to get rid of things like manual arming throttle source or change up where we talk about arming lockout? If a lot of that stuff doesn't make sense anymore, we should also change that up to reflect the new reality. It can be a different PR if you want, but it should happen before we make the redundant stuff go live.

vertiq-jordan:
> @vertiq-jordan General question, is any of the stuff on the existing arming pages outdated or inaccurate now? I recall that redundant throttle changes up some of the arming stuff. Do we need to update any of those pages to get rid of things like manual arming throttle source or change up where we talk about arming lockout? If a lot of that stuff doesn't make sense anymore, we should also change that up to reflect the new reality. It can be a different PR if you want, but it should happen before we make the redundant stuff go live.

Good call. I'd rather take care of it here. I'll do that now

docs/manual/manual_advanced_arming.rst
iq-fred, line 1: My only comment is about the stuff we talked on over slack, about allowing for info about the old stuff in some kind of form to be available for people as they transition

iq-fred (changes requested):

docs/manual/manual_advanced_arming.rst
```diff
@@ -13,6 +13,10 @@ simplifying flight controller integration.
 
 The :ref:`arming_module_support` section below details which Vertiq modules currently support advanced arming.
 
+.. note::
+    If your module does not support the :ref:`redundant throttle feature <redundant_throttle_manual>`, please also read the :ref:`legacy_arming_support` section for additional constraints 
+    on the advanced arming feature. You can find if your module's firmware supports redundant throttle on its family page.
```
iq-fred, line 18: @vertiq-jordan One last thing, this note is good and the legacy section is good, I may just sprinkle this note in a few more places. People rarely read our docs from the top to bottom in their entirety, so there's a good chance they might miss this up here. Dropping the note into a few other places where we removed stuff, like Always Armed, Armed State Transitions, Consecutive Arming Throttles, etc. may make sure people actually get that message

iq-fred (approved):

PR #130, opened 2025-03-10
update description of min and max for safety

vertiq-jordan:
We had a customer call out the fact that our descriptions on our PX4 min/max/disarmed configuration were wrong. This updates them, and gives further examples about what's going on.

Here's what they said, "On the PX4 documentation for PWM you have the disarmed us higher than the minimum us. https://iqmotion.readthedocs.io/en/latest/tutorials/pwm_control_flight_controller.html#px4-and-qgroundcontrol-configuration The recommendation from PX4 side last I read it was that the motor always spin at minimum setting – I suspect it has to do with maintaining control over the motor when a lightweight craft does aggressive manoeuvres. We have seen minimum saturation on some of our drones during testing. Having the motor stop spinning/disarm is not ideal in this case. I understand there is no right or wrong and the user is responsible for his own settings, but it might be good to align or reference the PX4 recommendations in the PX4 settings. I’ll see if I can find it in the current version of the PX4 guide if you are interested."

Built docs are here: https://iqmotion.readthedocs.io/en/feature-update_px4_min_max/tutorials/pwm_control_flight_controller.html#setting-px4-pwm-parameters


docs/tutorials/pwm_control_flight_controller.rst
```diff
@@ -461,10 +461,32 @@ Several parameters must be set properly to make sure the flight controller can c
...
+modules use a range of 1000us to 2000us for PWM throttle parsing, we can easily map our percentages to the correct Disarmed, Minimum, and Maximum values. When disarmed, we want the flight 
+controller to output a command that will disarm the module. So, we will set our Disarmed value to 1000us as it represents a 0% throttle. Now, we want to ensure the module arms when 
+the flight controller does, so we'll adjust our minimum flight controller output to fall inside of our :ref:`arming region <arming_throttle_regions>`. To ensure we fall inside of the 
+arming region, we will configure the flight controller to output a 2.5% throttle once armed. 2.5% of our 1000us range is 25, so we'll set our Minimum to 1025us. The maximum is 
+defined by the longest pulse Vertiq modules accept as PWM throttles, 2000us.
+
+.. warning::
+
+  It is strongly encouraged that your Minimum throttle output from the flight controller lies outside of your module's disarming region. If your module disarms while in flight, 
+  it can lead to vehicle failure. For the safest flight experience, your flight controller's outputs and your module's arming regions must align such that the flight controller will never
```
iq-fred, line 481: I wouldn't say "vehicle failure". Not saying its not true, I may just be a bit more specific, like "...it can cause your modules to stop responding to throttle commands until they re-arm".

docs/tutorials/pwm_control_flight_controller.rst
```diff
@@ -461,10 +461,32 @@ Several parameters must be set properly to make sure the flight controller can c
...
+
+.. warning::
+
+  It is strongly encouraged that your Minimum throttle output from the flight controller lies outside of your module's disarming region. If your module disarms while in flight, 
+  it can lead to vehicle failure. For the safest flight experience, your flight controller's outputs and your module's arming regions must align such that the flight controller will never 
+  disarm your module via throttles while the flight controller is armed.
+
+
+In the module configuration completed so far in this tutorial, there is no configured :ref:`disarming region <disarm_on_throttle>`. So, we will set both our Disarmed and Minimum outputs 
+to 990us in order to ensure that the module will not spin unexpectedly on 0% throttle outputs. The Maximum value can remain 2000us.
```
iq-fred, line 486: While you are completely right here that since there is no disarming region, this is totally safe if you are following all the instructions, I may be even more cautious because of how people so often tend to jump around and cherry pick from the documentation. If you did a setup similar to the one you described above, where the disarming is clearly lower than the minimum (1000 us vs. 1025 us, doesn't have to be those exact numbers if you don't want, but that idea) it may help drive home for the people who are just glancing at the pictures that disarming and minimum should probably be different most of the time. So I'm lobbying for changing it up here so that disarming is lower than minimum to illustrate this point more. We may not get quite all the way to 0% with minimum, but that's probably ok, if you let it go down really low like 1% or 2% or whatever, its pretty much stopped generally.

Something we can discuss, I just know that people often hop around and making them think a bit more about the difference between the disarming and minimum with our setup and pictures may be good.

iq-fred (changes requested):

docs/tutorials/pwm_control_flight_controller.rst
```diff
@@ -461,10 +461,32 @@ Several parameters must be set properly to make sure the flight controller can c
...
+
+.. warning::
+
+  It is strongly encouraged that your Minimum throttle output from the flight controller lies outside of your module's disarming region. If your module disarms while in flight, 
+  it can cause your modules to stop responding to throttle commands until they rearm. For the safest flight experience, your flight controller's outputs and your module's arming 
+  regions must align such that the flight controller will never disarm your module via throttles while the flight controller is armed.
+
+In the module configuration completed so far in this tutorial, there is no configured :ref:`disarming region <disarm_on_throttle>`. So, we will set our Disarmed output to 990us and our 
+Minimum outputs to 1000us in order to ensure that the module will not spin unexpectedly on 0% throttle outputs. Note that our Disarmed output should always be lower 
+than the Minimum output. The Maximum value can remain 2000us.
```
iq-fred, line 486: The thing about this is that unless you re-do the analog calibration to have 990us interpreted differently than 1000us, by default, then 990us and 1000us will both be interpreted as 0% throttles, so they are effectively the same. I know that in this example we have no disarming region, but I'm afraid people may get tripped up on not understanding that throttle interpretation thing if they use this as a reference for when they do try to do a disarming region. Any chance it could be like, 1000 us and 1005 us or something small like that? Just so they are really throttles that will actually be interpreted differently in the range? I know that means that the minimum will not be 0% anymore, but 0.5% or whatever is pretty darn close, and I know that functionally its not a meaningful difference with no disarming region

docs/tutorials/pwm_control_flight_controller.rst
```diff
@@ -461,10 +461,32 @@ Several parameters must be set properly to make sure the flight controller can c
...
+
+Once you have fully configured the output channel, you will see sliders appear under Actuator Testing. Please note that if your module is powered on and connected to the MAIN 1 output when it is enabled, 
+you will hear the module play its :ref:`two tone arming song <arming_song>` as PX4 begins transmitting Disarmed throttle commands (990us) immediately on channel activation. By default, Vertiq modules 
 take 0% throttle commands as arming throttles. 
 
 For more information on how to configure the module to properly interpret throttle commands, see the :ref:`manual_throttle` section of the Feature Reference
```
iq-fred, line 492: This is for further down: Should we update any of the DSHOT pictures as well to match the PWM stuff?

iq-fred (changes requested):

docs/tutorials/pwm_control_flight_controller.rst
vertiq-jordan: I don't think we need to, honestly. The DSHOT setup is independent of the PWM stuff, so I don't particularly care what's listed in the PWM boxes in the background

docs/tutorials/pwm_control_flight_controller.rst
```diff
@@ -461,10 +461,31 @@ Several parameters must be set properly to make sure the flight controller can c
...
+  regions must align such that the flight controller will never disarm your module via throttles while the flight controller is armed.
+
+In the module configuration completed so far in this tutorial, there is no configured :ref:`disarming region <disarm_on_throttle>`. So, we will set our Disarmed output to 1000us and our 
+Minimum output to 1005us. Note that our Disarmed output should always be lower than the Minimum output. The Maximum value can remain 2000us.
+
+Once you have fully configured the output channel, you will see sliders appear under Actuator Testing. Please note that if your module is powered on and connected to the MAIN 1 output when it is enabled, 
+you will hear the module play its :ref:`two tone arming song <arming_song>` as PX4 begins transmitting Disarmed throttle commands (990us) immediately on channel activation. By default, Vertiq modules 
 take 0% throttle commands as arming throttles. 
 
 For more information on how to configure the module to properly interpret throttle commands, see the :ref:`manual_throttle` section of the Feature Reference
```
iq-fred, line 491: Github won't let me comment on it, but the picture right below this where it shows 990 and 990 for Motor 1, is that not wrong now? Shouldn't we have a picture where it shows the 1000 and 1005 thing?

iq-fred (changes requested):

PR #138, opened 2025-04-14
Feature/update dronecan health info

vertiq-jordan:
Adds documentation about our new node health and status bit guys. Built here: https://iqmotion.readthedocs.io/en/feature-update_dronecan_health_info/communication_protocols/dronecan_protocol.html


docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
 The DroneCAN protocol requires that all nodes on a DroneCAN network periodically publish their status using the Node Status message. Vertiq modules support this behavior to conform 
 to the standard. A uavcan.protocol.NodeStatus message is published at 1 Hz during operation, reporting its health, current mode, and uptime.
 
-In normal operation a Vertiq module should always report its Health as OK, and its mode as Operational. No sub-modes or vendor specific status codes are currently supported.
+Vertiq's Node Health Indications
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Vertiq modules support all four standard DroneCAN `health indications <https://dronecan.github.io/Specification/7._List_of_standard_data_types/#:~:text=fatal%20malfunction.%0Auint2-,health,-%23%0A%23%20Current%20mode.%0A%23%0A%23%20Mode>`_. Modules enter each state given the following:
```
iq-fred, line 53: Very subjective, but somehow I feel weird about leaving off on the colon at the end and then having a bunch of separate headings. Maybe something like "The sections below describe the conditions under which modules enter each state" or some variant of that doesn't use a colon.

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
...
+
+Critical
+""""""""""
+#. The module's :ref:`total derate <derates>` reaches 0
+#. Your module's drive mode is in lockout due to :ref:`power safety <power_safety>`
+
+Warning
+""""""""""
+#. The module's supply voltage is greater than your drive's :ref:`v_max_start <brushless_drive>`. This typically indicates that the module is regenerating into a supply that cannot handle regenerated current, such as a typical benchtop power supply, and is limiting how much it regenerates to protect itself
```
iq-fred, line 62: Linking to the safety system stuff (https://iqmotion.readthedocs.io/en/feature-update_dronecan_health_info/manual/manual_safety_systems.html) might be better here than just the table, or linking both could be good to. The safety system part just gives a lot more context on what the heck this all really means for a customer. And maybe it would be better to call this Volts Limit Starting Voltage instead of v_max_start for a more descriptive name and to match what's in Control Center and on the Safety Systems page. The other links here got the safety system links, so it's consistent as well

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
 The DroneCAN protocol requires that all nodes on a DroneCAN network periodically publish their status using the Node Status message. Vertiq modules support this behavior to conform 
 to the standard. A uavcan.protocol.NodeStatus message is published at 1 Hz during operation, reporting its health, current mode, and uptime.
 
-In normal operation a Vertiq module should always report its Health as OK, and its mode as Operational. No sub-modes or vendor specific status codes are currently supported.
+Vertiq's Node Health Indications
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Vertiq modules support all four standard DroneCAN `health indications <https://dronecan.github.io/Specification/7._List_of_standard_data_types/#:~:text=fatal%20malfunction.%0Auint2-,health,-%23%0A%23%20Current%20mode.%0A%23%0A%23%20Mode>`_. Modules enter each state given the following:
+
+Critical
```
iq-fred, line 55: Possibly its obvious, but it may be good to specify that if any of these conditions are met its in this health state. It's potentially ambiguous right now if its all the conditions have to be true, or if just one of them. I think from context its probably obvious but it might be good to start off by saying something like "The module will enter Critical mode if any of the following conditions are true" or whatever. Something in each of these sections that makes it clear that you just need one to trigger it.

And that goes for all of these sections

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
...
+Healthy
+""""""""""
+When none of the conditions of Critical, Warning, or Error are met, the module is considered Healthy.
+
+
+Vertiq's Vendor Specific NodeStatus Code Breakdown
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Vertiq's vendor specific NodeStatus bits are used to indicate the state of various possible module errors. The bits are broken down as follows:
+
```
iq-fred, line 80: Any chance you could give us an example of what it would look like in a certain condition to illustrate your point? Like, pick some interesting arbitrary state, maybe our microcontroller is derating and our voltage is too high or whatever, and show what  you would expect the status code to be. Doesn't have to be that exact situation, but some interesting example. People really benefit a lot from examples, even if it seems obvious

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
...
+* Bits [0, 7]: Indicate the value of :ref:`Power Safety's fault_now <power_safety_client_params>` value. Please refer to the linked documentation for an additional breakdown of these bits
+* Bit 8: Indicates that the module's :ref:`speed derate <speed_derate>` is less than 1 but greater than 0. This means the motor is approaching an over-speed condition
+* Bit 9: Indicates that the module's :ref:`microcontroller temperature derate <microcontroller_derate>` is less than 1 but greater than 0. This means that the microcontroller's temperature is getting dangerously high
+* Bit 10: Indicates that the module's :ref:`coil temperature derate <coil_temp_derate>` is less than 1 but greater than 0. This means that the coil temperature is getting dangerously high
+* Bit 11: Indicates that the module's measured voltage is greater than your drive's :ref:`v_max_start <brushless_drive>`. This typically indicates that the module is regenerating into a supply that cannot handle regenerated current, such as a typical benchtop power supply, and is limiting how much it regenerates to protect itself
```
iq-fred, line 85: Same as above with maybe linking to the safety systems stuff on the voltage limits instead, more informative

docs/communication_protocols/dronecan_protocol.rst
```diff
@@ -47,7 +47,42 @@ uavcan.protocol.NodeStatus (Data Type ID = 341)
...
 
 Refer to the `uavcan.protocol.NodeStatus section of Standard Data Types <https://dronecan.github.io/Specification/7._List_of_standard_data_types/#nodestatus>`_ in the specification 
 for more details.
```
iq-fred, line 88: Maybe worth also saying that additional status information can be found in the ExtendedStatus guy and throwing a link to that, just so people who are trying to connect all of the status dots have a bit easier time.

iq-fred (changes requested):

iq-fred (approved):

PR #143, opened 2025-06-10
Add Buzzer Note to Slew Limit

vertiq-jordan:
Adding a note to the slew limit guy about how we don't have a separate buzzer/why the buzzer sounds different:

https://iqmotion.readthedocs.io/en/feature-add_buzzer_note_to_slew/manual/manual_safety_systems.html#slew-rate-limiter:~:text=require%20manual%20tuning.-,Note,-When%20adjusting%20your

### DIFF
--- a/docs/manual/manual_safety_systems.rst
+++ b/docs/manual/manual_safety_systems.rst
@@ -266,6 +266,12 @@ The following examples are meant to illustrate the effect that slew limiting has
 As you can see, the slew limit parameter helps protect against sudden changes in commanded voltage. This can be especially useful in :ref:`noisy/vibration heavy 
 environments <vibration_and_jittering>`. The amount of slew filtering that is necessary (if any) is completely application specific, and will require manual tuning.
 
+.. note:: 
+    
+    When adjusting your module's slew rate limit, it is possible that it may affect the module's :ref:`buzzer function <manual_buzzer_control>`. The module creates its buzzer tones by 
+    rapidly changing the motor's voltage, and not through any separate hardware (the modules do not have a separate buzzer). So, by lowering the slew limit specifically, you may 
+    cause your buzzer to sound quieter than expected as the module cannot generate the frequencies (changes in voltages) required to play a strong tone.
+
 Brushless Drive Current Limiter
 ====================================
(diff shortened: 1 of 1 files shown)


docs/manual/manual_safety_systems.rst
```diff
@@ -266,6 +266,12 @@ The following examples are meant to illustrate the effect that slew limiting has
 As you can see, the slew limit parameter helps protect against sudden changes in commanded voltage. This can be especially useful in :ref:`noisy/vibration heavy 
 environments <vibration_and_jittering>`. The amount of slew filtering that is necessary (if any) is completely application specific, and will require manual tuning.
```
iq-fred, line 268: I might move this up from the very bottom of this section so its a bit more visible. People may not actually make down this far.

Maybe put it right before the examples bit, e.g. before "The following examples..." to increase the chance of people seeing it

docs/manual/manual_safety_systems.rst
```diff
@@ -266,6 +266,12 @@ The following examples are meant to illustrate the effect that slew limiting has
...
 
+.. note:: 
+    
+    When adjusting your module's slew rate limit, it is possible that it may affect the module's :ref:`buzzer function <manual_buzzer_control>`. The module creates its buzzer tones by 
+    rapidly changing the motor's voltage, and not through any separate hardware (the modules do not have a separate buzzer). So, by lowering the slew limit specifically, you may 
+    cause your buzzer to sound quieter than expected as the module cannot generate the frequencies (changes in voltages) required to play a strong tone.
```
iq-fred, line 273: I'm sitting here thinking about if it has more to do with frequency or amplitude, because in my experience for a lot of the slew limit range it sounds like the right tone, just quiet, which feels like amplitude. But maybe if you get low enough in your slew limit the frequency might start being an issue too. But really it may not matter for customers and less is more might be better here. So maybe change the "generate the frequencies (change in voltages)" to just be "generate the changes in voltages" to spare us getting into the weeds of frequencies and amplitudes and such, just want them to understand that they are limiting the voltage we use for our buzzing too

iq-fred (changes requested):

iq-fred (approved):

PR #155, opened 2025-07-31
DSHOT Disarming Info

vertiq-jordan:
Found some missing documentation when writing the release blog. Adding it here: https://iqmotion.readthedocs.io/en/feature-add_info_about_dshot_disarming/manual/manual_advanced_arming.html

### DIFF
--- a/docs/manual/manual_advanced_arming.rst
+++ b/docs/manual/manual_advanced_arming.rst
@@ -332,7 +332,7 @@ DSHOT Disarm
 ##############
 The :ref:`DSHOT <hobby_dshot>` hobby protocol reserves one of its special values for commanding a disarm. Vertiq modules support this command. Whether or not the module will disarm on 
 reception of this message is determined by the value of the :ref:`allow_dshot_disarming_message <hobby_input>` parameter. If it is true, the module will perform a disarming transition on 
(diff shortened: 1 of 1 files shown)


docs/manual/manual_advanced_arming.rst
```diff
@@ -330,8 +330,12 @@ on the current armed state or manually transition the state as needed.
 
 DSHOT Disarm
 ##############
-The :ref:`DSHOT <hobby_dshot>` hobby protocol reserves one of its special values for commanding a disarm. Vertiq modules support this command, and will disarm if a DSHOT disarm 
-command is received.
+The :ref:`DSHOT <hobby_dshot>` hobby protocol reserves one of its special values for commanding a disarm. Vertiq modules support this command. Whether or not the module will disarm on 
+reception of this message is determined by the value of the :ref:`allow_dshot_disarming_message <hobby_input>` parameter. If it is true, the module will perform a disarming transition on 
+reception. If it is false, the disarming command is ignored.
```
iq-fred, line 335: Maybe worth mentioning that by default we usually have this as true so people don't think they need to go turn it on if they want this.

iq-fred (changes requested):

